PR #20, opened 2015-01-15
updated organizations data

sroosa:


sroosa:
waiting until it has ids.

sroosa:
oops. It actually was already updated with the ids. It's good to go.

PR #47, opened 2015-01-22
adding-new-vocab-maps

sroosa:
This updates the asset type map, and adds a title type and date type
map.


scripts/cleaner.rb
```diff
@@ -1,13 +1,15 @@
 require 'rexml/document'
 require 'set'
+require 'YAML'
```
afred: @mccalluc try lower case?

PR #122, opened 2015-02-17
87 deploy code to aws via opsworks

afred:


afred:
@mccalluc maybe wait on merging this just yet. I'm still not convinced opsworks isn't overkill.

afred:
^^ actually.. you could prob go ahead and merge it.. since it's likely to be the way we go regardless of deployment tool.

PR #214, opened 2015-03-10
updates about page text

sroosa:
replaces mission and vision statements (to be moved) with Alan’s
prospectus


mccalluc:
@sroosa : Looking at the test failure (the red "X"), it looks like you have unbalanced tags. See if the description of the problem here is helpful: https://travis-ci.org/WGBH/AAPB2/builds/53806106#L661

(and as always, just grab me if this isn't making sense.)

sroosa:
sorry about that. fixed all of the missing end tags.

PR #220, opened 2015-03-11
Org logos

mccalluc:
@sroosa : I'm moving the orgs to yaml. Let me know if you see any show stoppers here.


sroosa:
@mccalluc I just took a look at this, and I think it will work, but I have a question. Some of the paragraphs have colons in them, which seems to indicate something in the yml format, highlighting the proceeding string in blue. Should I edit the text so that it doesn't use any colons. That wouldn't be a big deal if it's the easiest solution.

PR #227, opened 2015-03-16
Completes-60-Images

sroosa:
30 images for each featured gallery


mccalluc:
@sroosa : I copied the new configs out of this branch and pasted them into #234, so just going to delete this.

PR #237, opened 2015-03-18
add-more-org-info

sroosa:
makes organization.yml completely up to date with info I've received from stations/contributing organizations.


sroosa:
@mccalluc Not sure what went wrong, but I assume it has to do with how i formatted the new data. Can we talk in person about the best way for me to edit this, especially to make sure I get the crucial indent right for the history and production information.

PR #241, opened 2015-03-19
Updates Gemfile.lock after running bundle update.

afred:


afred:
@mccalluc I think maybe something was getting screwed up with using relative paths from the `spec/` directory. Usually RSpec assumes you are running specs from at least one level down. Not sure what was going on under the hood, but moving `spec/code_style_spec.rb` to `spec/support/code_style_spec.rb` seems to have done the trick.

PR #280, opened 2015-03-27
added new station history and logo

sroosa:


mccalluc:
@sroosa : indenting needs to be spaces, not tabs, and I added a blank line between so that we'll get paragraph breaks.

PR #298, opened 2015-04-01
Opsworks deployment

afred:


mccalluc:
@afred : anything I should look at before merging?

afred:
@mccalluc hold off on this. I think i figured out how to do Capistrano with OpsWorks, and I don't think we''ll end up needing the `before_migrate.rb` hook. Also not sure if the Blacklight secret key thing will work.

PR #338, opened 2015-04-08
Link validation

mccalluc:
I'd want to expand it, but even this found some dead links. The logic for when it's run is a little confusing:
- never on travis (environment variable)
- by default, once a week for developers. (But you can force more frequently by deleting the watch file.)

Particularly if we're hitting remote sites, it can slow down the test suite substantially... but if it's something you have to remember to turn on, then it's not going to happen. This seems like a good balance.

Just scanning for local links right now, and already found a couple bad ones, so I think it's worth it.


afred:
@mccalluc looks good. Curious though, how did it catch the bad links in markdown?

mccalluc:
@afred : the checks are done after capybara has visited the page, so it really doesn't care how precisely it got rendered.

afred:
ah.. got it.

Issue #347, opened 2015-04-08
Site Map, or allowing Google to retrieve records in search results


caseyedavis12:
As a user, I would like to be able to conduct a search in Google and retrieve relevant item records on the AAPB website. @mccalluc says we can do this by creating a site map.

caseyedavis12:
We will wait on this until full reindex of solr.

caseyedavis12:
This is fixed! Example: https://www.google.com/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=captain+skillet+weasel

PR #350, opened 2015-04-08
Sitemap

mccalluc:
towards #347. When we have automatic reingests, we will get automatic sitemaps.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -7,3 +7,38 @@ Rails.application.load_tasks
 
 ZIP_URL = 'https://github.com/projectblacklight/blacklight-jetty/archive/v4.9.0.zip'
 require 'jettywrapper'
+require 'blacklight-sitemap'
+Rake::BlacklightSitemapTask.new do |sm|
+  # below are configuration options with their default values shown.
+
+  sm.resource_url = 'http://americanarchive.org/catalog'
+  sm.public_url = 'http://americanarchive.org'
+
+  # base filename given to generated sitemap files
+  # sm.base_filename = 'blacklight'
+
+  # Is the gzip commandline tool available? Then why not gzip up your sitemaps to
+  # save bandwidth?
+  sm.gzip = true
+
+  # for changefreq see http://sitemaps.org/protocol.php#changefreqdef
+  # valid values are: always, hourly, daily, weekly, monthly, yearly, never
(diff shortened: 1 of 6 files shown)


Rakefile
```diff
@@ -7,3 +7,38 @@ Rails.application.load_tasks
...
+  # Solr field used to retrieve from a document the value for the priority element for a url
+  # sm.priority_field = nil
+
+  # Solr query sort parameter
+  # sm.sort = '_docid_ asc'
+  
+  # Solr request handler. This can be useful when your Solr configuration already has
+  # a filter query appended.
+  sm.qt = nil
+end
```
afred: Generally, you'd define rake tasks in `lib/tasks/whatever.rake` using syntax like...
```ruby
namespace :blacklight do
  desc "something awesome"
  task :something_awesome do
    # code here
  end
end
```
Is there a reason for putting it straight into the Rakefile?

mccalluc: This is what the generator gave me out of the box, and I didn't know that it was un-idiomatic. Worth fixing?

scripts/download_clean_ingest.rb
```diff
@@ -173,10 +174,14 @@ def process()
...
       $LOG.info('Finished one big commit.')
     end
-
+    
     # TODO: Investigate whether optimization is worth it. Requires a lot of disk and time.
     # puts 'Ingest complete; Begin optimization...'
     # ingester.optimize
+    
+    $LOG.info('Starting sitemap generation...')
+    $LOG.info(`rake blacklight:sitemap`) # TODO: figure out the right way to do this.
```
afred: maybe tryi switching to

```ruby
Rake::Tasks['blacklight:sitemap'].invoke
```

Rakefile
afred: i don't know that it's not idiomatic either. Can clean it up later if Rakefile gets too messy.

scripts/download_clean_ingest.rb
mccalluc: I don't think I can do that unless I pull in a lot more of rake, and I'm not prepared to get in to a big refactoring right now. (At the very least, there will be conflicts with the PR that just came in.) I've filed #357 for the follow-up.

PR #351, opened 2015-04-08
fix #348

### LINKED ISSUE #348: Not all stations are appearing in the Organizations Facet
For some reason, not all stations are appearing in the Organizations facet. Two examples are Wisconsin Public Television and Wisconsin Public Radio.

mccalluc:


afred:
was there something else we needed to take facet limits off? I'm confused by @cbeer's suggestion.

PR #359, opened 2015-04-13
--query, --just-reindex, and rubocop cleanup

mccalluc:
@afred, please review.


scripts/lib/null_logger.rb
```diff
@@ -1,8 +1,9 @@
...
-  def initialize(*args)
+  def initialize(*_args)
   end
-  def add(*args, &block)
+
+  def add(*_args, &_block)
   end
-end
\ No newline at end of file
+end
```
afred: @mccalluc what is the point of `NullLogger`?

mccalluc: @afred : 
```
$ grep NullLogger -r scripts
scripts/lib/null_logger.rb:class NullLogger < Logger
scripts/lib/downloader.rb:    $LOG ||= NullLogger.new
scripts/lib/pb_core_ingester.rb:    $LOG ||= NullLogger.new
```

If these are being called from download_clean_ingest, $LOG will be set, but if it's being called from the unit tests, we actually don't need or want any output to the log. NullLogger does this.

mccalluc:
@afred: Anything else?

mccalluc:
Thanks! (Sorry for the random comment.)

PR #360, opened 2015-04-13
Tweaks for solr 5

mccalluc:
@afred: We can run under solr 5 with these tweaks.


mccalluc:
@afred : if you, could, these should be merged before I do other solr work.

afred:
Are these backward compatible with Solr 4.x?  I.e. the version that ships with Blacklight-jetty, and the one we're currently using for AAPB?

mccalluc:
It continues to work locally for me, and tests pass.

afred:
Hmm.. I'm starting to think twice. Note that the following are all true.
  * AAPB uses Blacklight
  * Blacklight user RSolr (as the HTTP client that talks to Solr)
  * Rsolr is known to work with Solr 4, not Solr 5

How wise is it to try to use Solr 5, knowing that the underlying stuff is only known to work with Solr 4?

cbeer:
What's wrong with RSolr and Solr 5?

afred:
@cbeer Unknown. I didn't see any versions-specific info in the RSolr docs, except that it seems to be written to work with Solr 4. Assuming that Solr 5 has introduced some breaking changes, we can't assume RSolr 1.x works with Solr 5.. right?

cbeer:
Did Solr 5 introduce breaking changes in the REST API?

afred:
I assumed so because semver. If they didn't, why the major release?

cbeer:
There were incompatible changes in the java API, and some deprecated behavior removed in the configurations, but I haven't seen any REST API changes. Or, as another data point:

https://github.com/projectblacklight/blacklight/pull/1100

afred:
Right on. Thanks for the input.

@mccalluc i think we're ok, given these assurances. At least if there are issues, we can leverage help from blacklight folks, who appear to be in the same boat :)

PR #369, opened 2015-04-28
Catch and log all WP API errors.

### LINKED ISSUE #368: SSL error with WP API
```
Faraday::SSLError in Home#show
Showing /Users/chuck_mccallum/starting/AAPB2/app/views/home/_blog.html.erb where line #4 raised:
SSL_connect returned=1 errno=0 state=SSLv3 read server certificate B: certificate verify failed
Extracted source (around line #4):
      <% Wordpress::Client.new.get_posts(host, number: 1).each do |post| %>
```
- There is a rescue right now, but it's too fussy. Fix that.
- and track down down the issue here... Is the WP certificate actually bad?
- and what caused this? Did certificates change just on my machine? That would explain why production is still up.

mccalluc:
fix #368

### DIFF
--- a/app/views/home/_blog.html.erb
+++ b/app/views/home/_blog.html.erb
@@ -14,7 +14,8 @@
       </div>
       
     <% end %>
-  <% rescue Faraday::ConnectionFailed %>
-    <!-- Connection to <%= host %> failed; Will try again when cache expires. -->
+  <% rescue => e %>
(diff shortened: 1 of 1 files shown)


afred:
waiting on Travis.. which seems to be having a rough time lately. @mccalluc maybe run tests locally for peace of mind, if you haven't already?

PR #372, opened 2015-04-28
There was not enough info to tell why it was not running on travis.

### LINKED ISSUE #367: Code style checks AWOL?
https://travis-ci.org/WGBH/AAPB2/builds/60275424#L592
>
code style
Checking **0** files for cruft...
  has no debug cruft
  has no merge cruft

Maybe paths on travis are different?

mccalluc:
Fixes #367, hopefully.

### DIFF
--- a/spec/support/code_style_spec.rb
+++ b/spec/support/code_style_spec.rb
@@ -1,25 +1,34 @@
 describe 'code style' do
   before :all do
-    @debugging = []
-    @merging = []
+    @debug = []
+    @merge = []
+    @error = []
 
-    # "vendor", in particular, is very large on Travis.
-    files_to_check = (Dir['{app,config,scripts,spec}/**/*.{e,}rb'] - [__FILE__.gsub(/.*\/spec/, 'spec')])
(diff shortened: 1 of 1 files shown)


spec/support/code_style_spec.rb
```diff
@@ -1,25 +1,34 @@
...
-    @merging = []
+    @debug = []
+    @merge = []
+    @error = []
 
-    # "vendor", in particular, is very large on Travis.
-    files_to_check = (Dir['{app,config,scripts,spec}/**/*.{e,}rb'] - [__FILE__.gsub(/.*\/spec/, 'spec')])
-    puts "Checking #{files_to_check.count} files for cruft..."
-    files_to_check.each do |path|
+    root = File.dirname(File.dirname(File.dirname(__FILE__)))
```
afred: would `Rails.root` be the same thing?

mccalluc: It would, except this test doesn't load `Rails`... but a test writer might expect it to be there.
-> See #374

Issue #343, opened 2015-04-08
Add Player


caseyedavis12:
When we start back on development in May, Karen would like to prioritize adding the Player to the records so that we can begin making content in Ci accessible asap.

mccalluc:
This has nothing to do with page download speed.

caseyedavis12:
Oops, I interpreted speed label as a higher priority item.
________________________________________

mccalluc:
@afred : drew, this blocks on having a way to deploy ci.yml to production server. Thoughts?

afred:
@mccalluc splain?

mccalluc:
@afred : to show the videos, it needs a ci.yml. I could just scp it to a server, but if there's a better way to do it as part deploy, that might be good to sort out. 

Thinking about it, I can make it fail gracefully (ie, just the status quo) if not configured, and the question of how to get the config out there can be handled separately.

afred:
Ah.. "ci" as in Sony Ci... i was thinking "continuous integration", and wondering what it had to do with anything.

Deploying production config is a use case that keeps coming up. Two strategies I've considered are...
  1. Keep production config in a private git repository (Stash) and write some custom capistrano tasks to pull it down and overwrite the defaults.
  1. Set production config values using environment variables that can be specified during the deployment.... e.g. `CI_API_KEY=xyz RAILS_SECRET_KEY=pdq [...] cap deploy`. Then you could have your YAML files use `ENV['CI_API_KEY'] || "some-default-value"`. Or.. you could write some capistrano tasks to build the actual files themselves from these values. We could even use capistrano's `ask` method to prompt for required environment variables, and fail the deployment if they are invalid.

Do either of those sound sane? Other ideas?

mccalluc:
Done. (Redeploy of config still needs some thought.)

PR #373, opened 2015-04-28
Player

mccalluc:
Towards #343: we still need ci.yml on production.


afred:
good work @mccalluc!

caseyedavis12:
Excellent! We are moving right along. Thank you, @mccalluc!

PR #374, opened 2015-04-28
Rails.root in all tests.

mccalluc:
Define `Rails.root`, even if we're not loading all of rails.


afred:
@mccalluc, this needs a rebase

mccalluc:
I think the final state is good (tests running,) but I don't think I rebased correctly.

afred:
hmm... appears to have brought in all the other comments from commits that were part of the rebase. This could be a github caveat. I just pulled the branch and it's history looks good. Merging.

PR #376, opened 2015-04-29
XML cleaning for tests. 

mccalluc:
done, hopefully. should not have gotten sucked into this.


mccalluc:
@afred : review?

afred:
it's hard for me to follow everything that's going on with the `ValidationHelper::expect_fuzzy_xml`. What is it doing that is different than validating against an xhtml xsd?

I'll go ahead and merge this, but could use some clarification on what's going on in case i need to dig into it some day.

mccalluc:
@afred: yeah, I really feel there must be a better way to do this, but I didn't find it. There are two key ways we are not xml:
- There are the HTML tags which don't need to be closed.
- There are attributes that don't have values.
In both cases, I just supply the missing part so that it parses as xml, but the attributes are the trickier one of the pair. Grab me at your convenience and we can walk through the RE.

PR #388, opened 2015-05-08
This works on FF and Chrome for me.

mccalluc:
@caseyedavis12 : This disables the context menu for me on Chrome and FF.
@afred : please review.


afred:
hey, that's simple enough! Supposedly this is supported in all browsers. Let's run test it out in browser stack to make sure after deployment.

PR #391, opened 2015-05-13
Exhibits

mccalluc:
@afred : This is coherent, but I can also add more to it.

### DIFF
--- a/app/assets/stylesheets/homepage.css
+++ b/app/assets/stylesheets/homepage.css
@@ -288,7 +288,8 @@
 
 	/* Home Jumbotron - Featured Collections */
 	#collections {
-		background-image:url(/home-bg/home-three.jpg);
+                /* TODO: a different, distinct image? */
+		background-image:url(/home-bg/home-one.jpg);
 	}
 
 	/* Home Jumbotron - Featured Collections */
--- a/app/models/exhibit.rb
+++ b/app/models/exhibit.rb
@@ -1,0 +1,39 @@
+require_relative '../../lib/markdowner'
+require 'yaml'
+
+class Exhibit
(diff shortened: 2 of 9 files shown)


afred:
would it be possible to get down to a single carousel template?

afred:
^^ let me rephrase that. Of course it's possible. But would it be a better design to have just one template for carousel that expects some params, and then set those param in the `HomeController`, from `Organization`, `Featured` or other models?

mccalluc:
Maybe renaming would help, but I think the partials themselves make sense right now:
`_featured_carousel.erb` and `_orgs_carousel.erb` both are wrappers for `_carousel.erb`:
- `_featured_carousel.erb` takes a `gallery` parameter, and a different value for the parameter is given on the home page and the search page.
- `_org_carousel.erb` doesn't take any parameters, since it is only used once, but it does supply optional parameters for `_carousel.erb` that make sure the "All Organizations" link works.

`_org_carousel.erb` could just be moved into its parent, but I had thought the parallelism of the `_*_carousel.erb`s was helpful... maybe move them to their own directory?

(I'm reluctant to pass parameters down from the controller: If there was some business logic which determined which gallery to show, then yes, but it feels like it's just a detail of the view.)

afred:
yeah, maybe it would be over engineering for this case. The wariness for me comes from the tight coupling of the models with the carousels. With this pattern, it looks like if we want to change the models that go into a carousel, or combine them in different ways, or order them in special ways, we're looking at writing another template, and possible another model, or helper, as opposed to changing some business logic in the controller.

But it's ok. Currently it's pretty low cost, and we probably won't have to do any of that what I just said :)

PR #412, opened 2015-05-15
Exhibits

mccalluc:
@afred : please review.

@sroosa : As soon as this is in, you could start making edits to exhibits.yml. It would need a redeploy and partial reingest for any changes to take effect. Front page display and sidebar facet are suppressed for the time being. Let me know (sooner than later) if this doesn't look like it will match your needs.

### DIFF
--- a/app/controllers/exhibits_controller.rb
+++ b/app/controllers/exhibits_controller.rb
@@ -6,7 +6,10 @@ class ExhibitsController < ApplicationController
 #  end
 
   def show
+    @all = Exhibit.all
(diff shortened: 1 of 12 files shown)


mccalluc:
reviewing failure...

afred:
looks like maybe you need to fix a fixture?

afred:
looks good. Interesting patterns emerging where we are using models that get their data from YAML files (YAML-base?). I wonder if we could pull the pattern out into a thin data-access module to include (or extend) in YAML-based models.

Issue #401, opened 2015-05-15
Genres filter


caseyedavis12:
In the genre filter, only show genres in the AAPB controlled vocabularies for genre. All other genres can be placed in a Miscellaneous genre filter.

sroosa:
So do you not want to other genres appearing in the records at all? So like on ingest, if it doesn't have an AAPB genre source, it should just replace it with "Miscellaneous?" Or do you want just in the facet?

caseyedavis12:
We could keep the other genres in the record. This is just for the facet.

mccalluc:
@sroosa : We need the vocab mapping file checked in?
@caseyedavis12 : While we could keep both the mapped and unmapped genre, we haven't done that for other mappings... are we sure this one needs to be a special case?

caseyedavis12:
Didn't we do this for Asset Type, except for the "Miscellaneous" request? If combining non-controlled vocabulary genres under "miscellaneous" in the genre facet would be too time-consuming, we could disregard this component of the request. We could instead only display controlled genres in the facet.

sroosa:
For Asset Type we changed the data after it was exported from the AMS and before it was ingested into solr, so it doesn't appear on the website. If we did it the same way, then only the Genres that are inline with our vocab would make it onto the website at all. Can we handle Genre the same way?

sroosa:
Talked with @caseyedavis12 offline and she said that she's happy to follow what we did for Asset Type. So we should be able to proceed that way.

mccalluc:
@sroosa:
Plain "Performance" comes up a lot: I'm going to add it to the list, but we can revisit that.

mccalluc:
PR merged. We will revisit to split genre(?)/topic(?), once we know what to call them. Closing.

PR #418, opened 2015-05-15
Genre map

mccalluc:
Towards #401, but needs reingest, at least.


app/views/home/_browse_list.html.erb
```diff
@@ -0,0 +1,74 @@
...
+            "War and Conflict", "Weather", "Women"].each do |term| %>
+              <li><a href="/catalog?q=<%= term %>"><%= term %></a></li>
+          <% end %>
+          
+        </ul>
+
+      </div>
+     
+    </div>
+</div>
```
afred: do we expect to modify these much? If so, we should do this in a way that automatically sorts the two lists, and evenly distributes the columns.

afred:
what is the reason for modifying the fixtures? Does this have to do with only wanting to display our sanctioned genres?

spec/fixtures/pbcore/clean-MOCK.xml
```diff
@@ -12,8 +12,8 @@
   <pbcoreSubject>explosions -- gratuitious</pbcoreSubject>
   <pbcoreSubject>musicals -- horror</pbcoreSubject>
   <pbcoreDescription>Best episode ever!</pbcoreDescription>
-  <pbcoreGenre>Horror</pbcoreGenre>
-  <pbcoreGenre>Musical</pbcoreGenre>
+  <pbcoreGenre/>
+  <pbcoreGenre>Music</pbcoreGenre>
```
mccalluc: @afred : Answering your fixture question, this fixture is modified because "Musical" and "Horror" are no longer approved genre terms, so you would never see them in the output from cleaning.

mccalluc:
Hold off on merging: need to fix the sorting (as you note above), and make it responsive.

afred:
10-4, ping me when it's ready. thanks!

mccalluc:
This is what it looks like:

screen shot 2015-05-22 at 11 35 57 am
screen shot 2015-05-22 at 11 35 03 am

- Dropping the Topic/Format distinction until we actually have that in the config.
- Element CSS tweak to get sections closer together in narrow layout.

TODO: 
- split "genre" into two vocab-maps, "format" and "topic", or something like that.
- corresponding changes at ingest, and in UI.

@afred: Good enough for now?

afred:
:+1:

caseyedavis12:
Looks awesome, thank you Chuck!!

PR #423, opened 2015-05-18
TOU click-through

### LINKED ISSUE #392: Click-through to view content in Online Reading Room
By October, a user will be required to click-through and acknowledge a terms and conditions (currently in review by Legal Counsel) prior to viewing or listening to the first video or audio item in their research session. This could be managed by cookies.

mccalluc:
except I think this may not be the right direction. 
- Will crawlers get confused? Should we try to sniff UAs?
- The other way to do it would be in the view, perhaps just blocking the player?
- If we did do it this way, with a separate controller, should assent be required for every page, or only those with media?
- Finally, is one button enough... or do we need a checkbox? or checkbox plus initials?
- Ok to apply this across the board right now, since we don't have any non-on-site access right now, anyway?


mccalluc:
(failures are expected, since you can't just go straight to a details page any more. Don't want to fix them unless we are sure this is the right direction.)

caseyedavis12:
If we did do it this way, with a separate controller, should assent be required for every page, or only those with media?

>> only those with media (records with a Ci ID)

Finally, is one button enough... or do we need a checkbox? or checkbox plus initials?

>>can you send me a screenshot of what it looks like?

mccalluc:
@caseyedavis12 : screen shot:
screen shot 2015-05-19 at 9 10 33 am

caseyedavis12:
Thank you. I will present the screenshot at our Rights meeting tomorrow to get counsel approval (to make sure that check box is sufficient). I am waiting for the Online Reading Room Rules of Use to be approved by the legal counsel, as well. This is different than the general terms and conditions on the website. The ORR Rules of Use will appear in the modal.

mccalluc:
@caseyedavis12 
- There is no checkbox here: If you want one, that's possible, but it's not here.
- To avoid confusion, this would not be called a modal in a web design sense: That would be a rectangle that appears over the normal page, with the background of the page greyed out, typically.

Neither a checkbox nor a modal is bad, but if we don't accurately describe the options in front of us things will get very confused.

caseyedavis12:
@mccalluc, the proposed solution below has been approved by the Legal Counsels. As soon as the Rules of Use has been approved, I'll deliver the text to you.

The legal counsels agreed that it should be presented to a user at the beginning of their research session. If they leave the website and come back to it, they should be presented with the Rules of Use again. Will that be difficult to implement?

[screen shot 2015-05-19 at 9 10 33 am]<https://cloud.githubusercontent.com/assets/730388/7703608/14682c66-fe07-11e4-8043-fed8120c5b06.png>

mccalluc:
[cookie](http://api.rubyonrails.org/classes/ActionDispatch/Cookies.html), with expiration set to an hour, updated with each request.

mccalluc:
**Do not merge** until we get the final language.

afred:
Oh.. so there are _two_ TOS the user has to click through? That's a drag.

afred:
:+1: go ahead an merge once language is final @mccalluc

caseyedavis12:
Hmmm. There should only be one "click through." There is the General Terms of Use (which is just a static page that the user can visit if they choose) and there is the ORR Rules of Use, which the use should have to click through to view video and audio — once every hour.

afred:
got it @caseyedavis12, thanks. @mccalluc explained it to me yesterday

app/controllers/catalog_controller.rb
```diff
@@ -169,15 +171,33 @@ def show
...
+    
+  end
+  
+  private
+  
+  def requires_eula?
+    !session[:affirm_terms] && 
+            (@pbcore.video? || @pbcore.audio?) &&
+            !bot? &&
+            GeoIPCountry.instance.country_code(request.remote_ip) == 'US'
```
afred: @mccalluc we don't require agreement to Terms for people outside the U.S.?

mccalluc: No, because they're entirely blocked from viewing the video... but the spaghetti logic does worry me.

afred: where do we block non-US viewers?

afred:
@caseyedavis12 are we still waiting for final language on this?

caseyedavis12:
@afred: yes, we are still waiting for final language.

mccalluc:
@caseyedavis12 : still waiting?

caseyedavis12:
@Mccalluc @afred Yes. Do not worry, I will let you know as soon as it is finalized. This is top priority for me.

mccalluc:
Made new PR #573 by rebasing for clarity. Closing this PR.

PR #428, opened 2015-05-21
updates asset type vocab map

sroosa:
with new term added to asset type controlled vocab.


mccalluc:
Looks good. (Previous version had DOS line endings, for some reason?)

PR #431, opened 2015-05-22
Custom 404

### LINKED ISSUE #161: custom 404?
Do we need a custom 404 page? Should it share the header and footer, or be its own thing?

mccalluc:
Would fix #161. Not super-confident that this is the right approach.


afred:
:+1: looks good

Issue #396, opened 2015-05-15
Capitalize Titles


caseyedavis12:
@sroosa: can we capitalize titles of all items in the catalog

caseyedavis12:
@mccalluc: look into doing this programmatically. Perhaps ingest time cleanup. Call letters will need to be disregarded because we want those capitalized.

mccalluc:
I'm going to close this: PR is merged... but note that nothing really changes until there's a reingest. If you want to reopen/refile specifically for that, feel free.

PR #434, opened 2015-05-26
Title casing on ingest.

mccalluc:
- Do we have examples of actual records that need this kind of fix?
- This is applied only to titles which are all lower case... maybe that's not right?
- Capitalization logic itself is questionable.
- Do we have examples where call letter special casing is necessary? How do we distinguish "WILL" from "will"?

### DIFF
--- a/scripts/lib/cleaner.rb
+++ b/scripts/lib/cleaner.rb
@@ -77,12 +77,22 @@ class Cleaner # rubocop:disable Metrics/ClassLength
     }
     
     match_no_report(doc, '/pbcoreTitle') { |node|
-      if node.text !~ /[A-Z]/ && node.text =~ /[a-z]/
-        node.text = node.text
-          .gsub(/\b\w/) { |match| 
+      if (node.text !~ /[A-Z]/ || node.text !~ /[a-z]/) && node.text =~ /[a-zA-Z]/
+        # ie, either all upper or all lower, and it has letters.
+        node.text = node.text.downcase
+          .gsub(/\b\w/) { |match|
+            # First letters
             match.upcase 
           }
-          .gsub(/\ba|an|the|and|but|or|for|nor|yet|as|at|by|for|in|of|on|to|from\b/i) { |match|
+          .gsub(/\b(NPR|PBS|APTS|APT|APM|PRX|NETA|ITVS|CPB)\b/i) { |match|
+            # Known acronyms
+            match.upcase
+          }
+          .gsub(/\b[^AEIOUY]+\b/i) { |match|
+            # Unknown acronyms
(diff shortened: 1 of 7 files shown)


mccalluc:
@caseyedavis12 / @afred : feedback appreciated.

caseyedavis12:
@mccalluc: there are also a lot of records that are in all caps. We would like for those to be title case as well. Example: http://americanarchive.org/catalog/cpb-aacip_16-jw86h4d707

An example of "WILL" call letters in a title is: http://americanarchive.org/catalog/cpb-aacip_16-79v15q57

Other examples where special casing is necessary is anytime the call letters for a station appear in the title. Can you add special logic to overlook call letters for participating stations, as well as NPR, PBS, APTS, APT, NET, APM, PRX, NETA, ITVS, and CPB?

mccalluc:
(Towards #396)

mccalluc:
Note that there is no special processing for "WILL"
- This code is only an ad-hoc clean-up when a title doesn't contain both upper and lower case.
- I think that even in this corpus, use as call letters will be less common that use as a word.
- "US", and "LA", among others, will be cleaned up, but it's easy to construct examples where that is not the right course: I might want to remove them from the huge RE as well.

scripts/lib/cleaner.rb
```diff
@@ -75,6 +75,48 @@ def clean(dirty_xml, name='not specified')
...
+              |WMEB|WNED|WNET|WNYC|WOJB|WOSU|WQED|WQEJ|WRFA|WRNI|WSIU
+              |WTIP|WTIU|WUFT|WUMB|WUNC|WUSF|WVIA|WVIZ|WWOZ|WXXI|WYCC
+              |WYSO|WYSU|YSU)\b/xi) { |match|
+            # Based on:
+            #   ruby -ne '$_.match(/[A-Z]{2,}/){|m| puts m}' config/organizations.yml \
+            #     | grep '[AEIOUY]' | sort | uniq | ruby -ne 'print $_.chop; print "|"'
+            # Removed: AM, AMBER, COLORES, COSI, FETCH, NET, RISE, SAM, STEM, TOLD, WILL
+            match.upcase
+          }
+          .gsub(/\b[^AEIOUY]+\b/i) { |match|
```
afred: Does this match any word that does not have a capital vowel? What's the logic here?

mccalluc: The `i` flag makes it case insensitive, but yeah, if there's a word without vowels, it's probably an acronym, and it should be all-caps.

afred: ah.. missed the `i` flag. Got it.

PR #436, opened 2015-05-28
Get it to use the new gem. 

mccalluc:
ALL tests pass, including the ones that are not run on travis. (There had been some rot.)

@afred, please review this, and the new gem: https://github.com/WGBH/sony-ci-api


afred:
This looks great Chuck, well done!

I would go ahead and tag the version at 0.1.0. Then allow the middle digit to increase as features are added, and reserve the least significant digit for bug fixes within a feature set. I think that is kosher semver.

I'm guessing releasing this under MIT will be fine with folks at Sony? Thoughts?

Next up, hook up Travis?

mccalluc:
@afred:
> I'm guessing releasing this under MIT will be fine with folks at Sony? Thoughts?

Before we advertise it to them, I'd want to clean up some of the things which are embarrassing. Since we're essentially doing work for them (*"We are actively working on building out libraries to make working with Ci even easier. Drop a line to our Customer Service and let them know your prefered language to help us prioritize them."*), I really hope there's nothing that would bother anyone here.

> Next up, hook up Travis?

Not sure there's a good way to do that: the tests are actually making API calls against Ci. So, if I checked in the ci.yml, we could have it run on travis, but then my password and keys would be published to the world.

That said, another complementary approach would be to test that the HTTP requests are what they should be, without actually hitting a server... Are you thinking that would be good?

afred:
Yes. The idea would be to test the edges of the gem's api without requiring the 3rd party thing that it's communicating with. For outgoing messages to Sony Ci, we can just check the expected message. For incoming message, that would require mocking Sony Ci responses -- which we can do if we're feeling ambitious. But keep in mind that mocks of 3rd party stuff would have to keep in step with the real deal, otherwise you'll end up code that passes tests but is actually broken. So we'd only want to go down that path if we deemed it worthwhile.

PR #442, opened 2015-06-08
html_safe to prevent double encoding

### LINKED ISSUE #424: symbols in "from the blog" on homepage
In the “From the Blog” section of the website, the new headline reads:
 
Accessing Historic Public Media: A Researcher&#8217;s Experience

 
Can you change the ```&#8217;``` to an apostrophe?

mccalluc:
fixes #424

### DIFF
--- a/app/views/home/_blog.html.erb
+++ b/app/views/home/_blog.html.erb
@@ -3,7 +3,11 @@
   <% begin %>
     <% Wordpress::Client.new.get_posts(host, number: 1).each do |post| %>
       <h6>From the Blog</h6>
-      <h3 class="list-group-item-heading"><a href="<%= post.url %>"  target="_blank" title="From the AAPB Blog:<%= post.title %>"><%= post.title %></a></h3>
+      <h3 class="list-group-item-heading">
+        <a href="<%= post.url %>"  target="_blank">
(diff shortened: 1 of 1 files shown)


app/views/home/_blog.html.erb
```diff
@@ -3,7 +3,11 @@
   <% begin %>
     <% Wordpress::Client.new.get_posts(host, number: 1).each do |post| %>
       <h6>From the Blog</h6>
-      <h3 class="list-group-item-heading"><a href="<%= post.url %>"  target="_blank" title="From the AAPB Blog:<%= post.title %>"><%= post.title %></a></h3>
+      <h3 class="list-group-item-heading">
+        <a href="<%= post.url %>"  target="_blank">
+          <%= post.title.html_safe %>
+        </a>
+      </h3>
```
afred: Did you drop the `title` attribute on the `<a>` tag intentionally?

mccalluc: Yeah: I couldn't see that it added anything, but no strong objection if you think it is useful.

afred: i was thinking it was good for accessibility, but after some reading I don't think that's the case. Merging.

PR #449, opened 2015-06-11
added kcpt info

sroosa:


afred:
I grew up on KCPT! Sesame Street, Mr Rogers and Bob Ross ftw!

PR #451, opened 2015-06-12
org config fixes

mccalluc:
@sroosa: remember that you need a blank line in between to get paragraphs.

Also, line wrapping will make diffs easier to read.


afred:
thanks @mccalluc

PR #453, opened 2015-06-12
Open graph

mccalluc:


app/views/layouts/blacklight.html.erb
```diff
@@ -0,0 +1,55 @@
+<!DOCTYPE html>
+<html lang="en" class="no-js" prefix="og: http://ogp.me/ns#">
```
mccalluc: @afred: I only want to add the prefix element. Do you think blacklight maintainers would be interested in a PR just to make a partial for attributes on the `<html>` element?

afred: maybe. Are you thinking a partial that could be overwritten, or config'd from a yaml file or something?

afred: ah.. i see what you're doing now. Which is very cool btw. Yeah, i think that would be a good change for BL in general. If you want to get a feel for acceptance before submitting a PR, I'd ask the IRC channel #projectblacklight

app/views/catalog/show.erb
```diff
@@ -1,3 +1,38 @@
+<% content_for :head do %>
+  <meta property="og:title" content="<%= @pbcore.title %>" />
+  <meta property="og:url" content="http://americanarchive.org/catalog/<%= @pbcore.id %>" />
+  <meta property="og:type" content="<%=
+    # See http://ogp.me/#types
+    # The OG type vocabulary is impoverished.
+    if @pbcore.video?
+      'video'
```
afred: should this be one of `video.movie`, `video.episode`, `video.tv_show`, `video.other`? It's unclear to me from the docs whether using just `video` would work.

app/views/layouts/blacklight.html.erb
mccalluc: (Filed https://github.com/projectblacklight/blacklight/issues/1220: Not sure what is best so thought that might be a better place.)

afred: looks great. I'll merge this in the meantime. Interested to see the effects. Do you have a good way to demo? Screen grabs of FB sharing, etc?

PR #456, opened 2015-06-16
Exhibit tweaks 432 433

### LINKED ISSUE #432: Proposal: provide some context when viewing exhibit results?
@caseyedavis12 : Thinking some more about the problem of providing some context when viewing an exhibit result set. What I think might be possible would be to supplement the list of facets at the top.
screen shot 2015-05-22 at 1 24 35 pm

The text there could be the first *x* characters in the description, or we could create a new field just for this.

I'm not super enthusiastic about this, but it probably wouldn't be that hard to do, either.

### LINKED ISSUE #433: Proposal: provide context on individual items in exhibits
screen shot 2015-05-22 at 1 43 37 pm

@caseyedavis12 : This is probably a bad idea, but wanted to put it out there. Instead of just supplying a list of IDs, it could be a hash, with the value of each pair being an in-context description. Just as we add exhibit membership information on ingest, we could add more description to individual records. (Probably record it as another description, with an identifying attribute.) Then it could be shown in the results list, and/or on individual records.

(Individual record pages are state-less: It would be a really bad idea to show it differently, depending on how you got to it.)

mccalluc:
fix #432
fix #433

### DIFF
--- a/app/models/exhibit.rb
+++ b/app/models/exhibit.rb
@@ -4,6 +4,7 @@ require 'yaml'
 class Exhibit
   attr_reader :slug
   attr_reader :name
+  attr_reader :items
   attr_reader :ids
   attr_reader :thumbnail_url
   attr_reader :html
@@ -33,7 +34,8 @@ class Exhibit
   def initialize(hash)
     @slug = pop(hash, 'slug')
     @name = pop(hash, 'name')
-    @ids = pop(hash, 'ids')
+    @items = pop(hash, 'items')
+    @ids = @items.keys
     @thumbnail_url = pop(hash, 'thumbnail_url')
     @html = Markdowner.render(pop(hash, 'md'))
     fail("unexpected #{hash}") unless hash == {}
--- a/app/models/exhibit.rb
+++ b/app/models/exhibit.rb
@@ -9,7 +9,11 @@ class Exhibit
   attr_reader :html
   
   def self.find_by_slug(slug)
-    @@exhibits[slug]
+    @@exhibits_by_slug[slug]
+  end
+  
+  def self.find_by_name(name)
+    @@exhibits_by_name[name]
   end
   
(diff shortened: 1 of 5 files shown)


foo4thought:
@afred : Chuck and I looked over this and I voted to *not* merge at this time because you ought to see it first.

app/models/exhibit.rb
```diff
@@ -29,18 +34,28 @@ def pop(hash, key)
...
   
+  @@exhibits_by_name = Hash[
+    YAML.load_file(Rails.root + 'config/exhibits.yml').map do |hash|
+      [hash['name'], Exhibit.new(hash)]
+    end
+  ]
+  
   @@exhibits_by_item_id = Hash[
     Exhibit.all.map{ |exhibit| exhibit.ids }.flatten.uniq.map do |id|
       [id, Exhibit.all.select { |exhibit| exhibit.ids.include?(id) } ]
```
afred: @mccalluc does `exhibit.ids` still exist? It looks like you removed it in the YAML file below.

mccalluc: So, in order to let you cram a little description on each one, I changed it from a list of IDs to a hash. The `ids` method still works: it is just getting the keys.

(I was on the fence about even the idea of this per-item adhoc metadata; I wouldn't fight hard for it if you don't think it's the right direction.)

afred: ah.. got it. I'm ok with this solution. I prefer it to sticking exhibit-specific-metadata in the AMS.

I just missed line #'s [8](https://github.com/WGBH/AAPB2/blob/exhibit-tweaks-432-433/app/models/exhibit.rb#L8) and [38](https://github.com/WGBH/AAPB2/blob/exhibit-tweaks-432-433/app/models/exhibit.rb#L38).

PR #461, opened 2015-06-22
Turn off referer check. Fix #459.

### LINKED ISSUE #459: Turn off referer check inside WGBH
Some WGBH folks are allowed to download. Knowing the super-secret URL is considered sufficient security for this internal case. Just need to turn off the referer check in the media controller for WGBH users.

mccalluc:
I would like this to be tested, but it uses Ci... mock?
@foo4thought : please review.


foo4thought:
Looks OK to me.  Will be restored later?

PR #463, opened 2015-06-22
Use the html scrubber on view. fix #439, for now.

### LINKED ISSUE #439: Proposal: strip HTML on ingest?
There are a fair number of records with raw HTML: If the ugliness of the data is a concern, maybe we should look into this?

(There's a lot of overlap between these sets: the numbers can't be added.)
- 5724: http://americanarchive.org/catalog?q=href
- 5294: http://americanarchive.org/catalog?q=div
- 7248: http://americanarchive.org/catalog?q=nbsp
- 11172: http://americanarchive.org/catalog?q=quot ("quote" stems down to this, but if the character entity shows up in the description, it probably shows up multiple times, so TF-IDF ranks it highly.)

We have lots of straight-up html, but also cases where it looks like a half-assed cleanup was done at some point: the angle-brackets might be missing but the tag names are still in place, for example.

options:
- Status quo? Maybe the links there will be useful to someone.
- Try to preserve HTML semantics / white listing. (Would really advise against this.)
- Render to text at ingest
- ... and maybe preserve `<p>` or `<div>` as line breaks in resulting text?
- Render AND do some adhoc regexes to clean up the stuff that can't render?
- Or render to text at render time? Small overhead on each page view; wouldn't require reingest; I'd feel uncomfortable with this as a long term fix, but as a trial to see if we like the result, it's not a bad approach.

(I've just been looking at the descriptions, though there are some of the same problems in [other fields](http://americanarchive.org/catalog?search_field=titles&q=nbsp) as well.)

mccalluc:
@foo4thought : review.

### DIFF
--- a/app/models/pb_core.rb
+++ b/app/models/pb_core.rb
@@ -2,6 +2,7 @@ require 'rexml/document'
 require 'rexml/xpath'
 require 'solrizer'
 require_relative 'exhibit'
+require_relative '../../lib/html_scrubber'
 
 require_relative 'organization'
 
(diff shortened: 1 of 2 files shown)


mccalluc:
hmmm... not able to reproduce the failure locally.

PR #466, opened 2015-06-24
Genre format 427

mccalluc:
@foo4thought : please review.


mccalluc:
*WARNING*: This introduces a new facet; We'll need to do a full reindex after merging this or there will be dead links. Alternatively, those parts could be commented out, and leave it for a future PR to uncomment them. Thoughts?

afred:
@caseyedavis12 Do you need to show this feature to anyone in the next two weeks? If not, then I'd prefer to hold off on merging until I finish some of the deployment workflow tasks I have assigned to me.

If you _do_ need it sooner than that, then we can add some patch-work to get it working using current deployment methods.

caseyedavis12:
@afred: I do not need to show this feature to anyone in the next two weeks.

mccalluc:
@afred: should be ready to go, now.

PR #477, opened 2015-06-29
Make titles 100x more important in text search.

### LINKED ISSUE #465: Weight titles more heavily in search
There are some cases where the default relevance sort would be better if it weighted titles more heavily. (As it is, the title just goes in the bag of words along with the description and everything else.)

TODO: We need to identify more examples: For "Civil Rights", the WGBH series of the same name might come higher in the results.

mccalluc:
It does move the title which includes the search term up the list.
TODO: Is 100x right? Could it be over-weighted? Gather real examples.
Fixes #465.

### DIFF
--- a/jetty/solr/blacklight-core/conf/solrconfig.xml
+++ b/jetty/solr/blacklight-core/conf/solrconfig.xml
@@ -86,8 +86,8 @@
             http://wiki.apache.org/solr/LocalParams
        -->
 
-       <str name="qf">text</str>
-       <str name="pf">text</str>
+       <str name="qf">text titles^100</str>
+       <str name="pf">text titles^100</str>
        
        <str name="titles_qf">titles</str>
        <str name="titles_pf">titles</str>
--- a/jetty/solr/blacklight-core/conf/solrconfig.xml
+++ b/jetty/solr/blacklight-core/conf/solrconfig.xml
@@ -86,8 +86,8 @@
             http://wiki.apache.org/solr/LocalParams
        -->
 
-       <str name="qf">text titles^100</str>
-       <str name="pf">text titles^100</str>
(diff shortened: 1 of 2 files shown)


afred:
@mccalluc i'm not really sure if 100x actually means in this context.

Suppose I have 3 documents titled:
  * "John F Kennedy" (about the president)
  * "Dead Kennedys" (about the band)
  * "Bay of Pigs" (about the incident)

If I search for "John Kennedy", I would expect the following order:
  * "John F Kennedy"
  * "Bay of Pigs"
  * "Dead Kennedys"

Would heavier weight on title produce this result? Would we have to add weight to other fields (e.g. subject/topic/genre et al)?

In any event.. i'll go ahead and merge this and we can see what happens.... unless there are any objections... @caseyedavis12?

mccalluc:
So you know how solr relevance retrieval treats the document as a bag of words and does TF IDF? This dumps a hundred copies of the title words into the bag, giving you a much bigger TF.

Or something like that.

afred:
is 100x a good place to start, instead of 2x, 5x, 10x, etc? Also.. any other fields we want to boost?

I'm down to experiment, but I want to avoid introducing bugs that may be slow to discover if we're only relying on spot checking a few test queries. Might be nice to have some contrived fixtures that would respond to fine tuning.

Also, @jcoyne [pointed this out](https://github.com/projectblacklight/blacklight/wiki/Extending-or-Modifying-Blacklight-Search-Behavior#extending-blacklightsearchhelpersearch_params_logic) as a way to set boosts at query time.

mccalluc:
@afred: This could be deployed without reindex, but it is a schema change, if we'd like to get it in...

PR #480, opened 2015-06-29
Multi facet 377

### LINKED ISSUE #377: Discuss: Facet behavior can be misleading for fields that are usually single-valued.
@mccalluc @sroosa @caseyedavis12

For instance, the "Organization" facet. Here's an exercise that highlights the confusion:
  1. Do a search for "marathon"
  1. Open the "Organization" facets and notice that KCAW has 16 results, and KCTS has 33 results.
  1. Now click on KCAW. Notice that KCTS now has 0 results.

This is because Solr is calculating the number of records that have _both_ KCAW **AND** KCTS.

This method of faceting works great when filtering on multiple fields. For instance, if I filter on "Organization" and "Asset Type" I would expect the search to return results that match _both_ of my criteria, using **AND** (as opposed to matching _either_ of my criteria, using **OR**).

It also makes sense for truly multi-valued fields, like Genre, which an have any number of values for any single record. When I filter on "News" and "Music", I would expect to get results that satisfy _both_ of those genres.

It gets less user-friendly though, when we have a field like "Organization" or "Media Type". The common case for these fields is to only have one value -- e.g. the "Media Type" is _either_ "Moving Image" **OR** "Sound", but rarely both (if ever).

It might be worth it to see what happens if we switch to using single-valued fields for those fields that have single values most of the time. I think these would include "Organization", "Media Type", "Asset Type", and "Access".

I have not seen how Blacklight deals with this, so it would be an experiment, but it might make the facets more intuitive and useful.

I'm guessing we defaulted to multi-valued fields everywhere to stay flexible, and be able to handle those rare cases when, for instance, a record legitimately has 2 organizations. But if we can get better UI by handling the edge cases a little differently, it could be worth it.

mccalluc:
@afred: could you take a look at this approach to fix #377... Not sure how close this is to what we want.

### DIFF
--- a/config/environment.rb
+++ b/config/environment.rb
@@ -4,6 +4,7 @@ require File.expand_path('../application', __FILE__)
 # Initialize the Rails application.
 Rails.application.initialize!
 
+# Monkey-patches which might be made into PRs?
 
 module Blacklight::Solr
   class SearchBuilder < Blacklight::SearchBuilder
@@ -15,7 +16,7 @@ module Blacklight::Solr
 
       prefix = ""
       prefix = "{!#{local_params.join(" ")}}" unless local_params.empty?
-
+      
       fq = case
         when (facet_config and facet_config.query)
           facet_config.query[value][:fq]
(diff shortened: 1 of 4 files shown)


mccalluc:
For reference: screen shot 2015-06-30 at 9 24 10 am

The translation of the url parameters to solr takes place on a parameter-by-parameter basis, and for ORing to work, the facets to be combined need to be in a single solr term... and the easiest way to do that is to put them all in a single url parameter. In the abstract, it would be nice to decouple the front and the back, but that would take us farther away from the Blacklight code base.

config/environment.rb
```diff
@@ -3,3 +3,102 @@
...
+             (value.is_a?(Float) or (value.to_f.to_s == value if value.respond_to? :to_f))
+          "#{prefix}#{facet_field}:#{RSolr.solr_escape(value.to_s)}"
+        when value.is_a?(Range)
+          "#{prefix}#{facet_field}:[#{value.first} TO #{value.last}]"
+        else
+          # Before monkey-patching:
+          #"{!raw f=#{facet_field}#{(" " + local_params.join(" ")) unless local_params.empty?}}#{value}"
+          # Do all the extra parts matter?
+          prefix + value.split(/ OR /i).map {|single| "#{facet_field}:\"#{single}\""}.join(' ')
+          # END patch
```
afred: This changes the output quite a bit. I'm not solr-savvy enough to know if it's logically equivalent. E.g.
```ruby

# assuming this...
facet_field, value, local_params = "title", "The Illiad", %w(paramOne paramTwo)
prefix = "{!#{local_params.join(" ")}}" unless local_params.empty?

# The old way...
"{!raw f=#{facet_field}#{(" " + local_params.join(" ")) unless local_params.empty?}}#{value}"
# => "{!raw f=title paramOne paramTwo}The Illiad"

# The new way
prefix + value.split(/ OR /i).map {|single| "#{facet_field}:\"#{single}\""}.join(' ')
# => "{!paramOne paramTwo}title:\"The Illiad\"" 

```

Does the new query behave the same as the old query in the absence of " OR "?

afred:
@mccalluc did you look yet at `blacklight_advanced_search`, that Eben referred to in the blacklight-development thread? Here's his example: https://www.digitalcommonwealth.org/advanced.

I think this gets at the idea pretty well. I think it covers all of the functionality we're looking for.

Not crazy about the UX though.

The more I think about my own search behavior (which I extrapolate to the universe of web users, naturally), it seems that the default facet behavior is actually the _least_ desirable. Only rarely would I want to select two values from the same facet and get the intersection of search results, as opposed to the union.

mccalluc:
@afred: I looked when I began, but it seemed overly complicated then, but knowing how things work now, it might seem better. I'll come back to this.

mccalluc:
@afred: How do you feel about this right now? Even assuming the machinery is right, am I applying it to the right facets?

afred:
needs a rebase, and I am having a hard time following the tests in `spec/features/catalog_spec.rb`. I think it will be clearer if i run them locally. Have to split for today though.

PR #486, opened 2015-06-30
Link to advanced search

mccalluc:
- Making it an "a" would require more CSS tweaks, but maybe that would be better?
- Does not preserve entered search: if there are any search terms, then we don't get the advanced page. Is that ok? We could set up an actual, separate page at "/advanced", but I'm not sure that's warranted.
- Same thing for smaller search at top?

This PR can be handled independently of a PR for the search page itself.

For reference:
screen shot 2015-06-30 at 1 23 13 pm


caseyedavis12:
*   We can ask Mike to add making this an "a" to his list of tasks for late July. Is an "a" the most intuitive way to indicate that an advanced search is an option?
  *   I think it's fine that it does not preserve the entered search.
  *   We're going to use /catalog for the advanced search page, right?
  *   Same thing for smaller search at the top.

PR #487, opened 2015-06-30
Advanced search UI

mccalluc:
Needs design work, and js is kludgy, but is this the functionality we want? @caseyedavis12 ?

One problem is that if someone selects a facet it will just do that, ignoring any text search.

screen shot 2015-06-30 at 3 07 59 pm

### DIFF
--- a/app/views/catalog/_home_text.html.erb
+++ b/app/views/catalog/_home_text.html.erb
@@ -1,5 +1,5 @@
 
-<form action="/catalog">
+<form action="/catalog" id="advanced">
   <input name="q" type="hidden"></input>
   
   <label>Search</label><input id="include"></input>
@@ -18,7 +18,7 @@
 
 <script>
   $(function(){
-    $('form').change(function(){
+    $('#advanced').change(function(){
       var include = $('#include').val();
       var exclude = $('#exclude').val();
       var q = ( $('#exact').prop('checked') ? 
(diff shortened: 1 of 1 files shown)


caseyedavis12:
Thanks Chuck!

1. Can we add info/interface to indicate that the user can do AND or OR boolean searches?

2. Can we let them search within a date (year) range?

See here for an ideal functionality: http://libcat.csd.sc.edu:81/search/X

mccalluc:
@caseyedavis12 : https://www.digitalcommonwealth.org/advanced might be a better example to start with, if we want to investigate the advanced search plugin. There are some things we might do differently:

- Since we only have "all text", "title", and "about" for searching, I would just give each its own field, instead of all the selectors.
- What facet checkboxes do we need?
- I find the fine print over in the sidebar overwhelming.

caseyedavis12:
@mccalluc: I am inclined to include the field selectors, because this will allow people to see that they can conduct AND searches within the same field. Is there a better way to do this than provide the selectors?


  *   On the right side of the search boxes, can we include selectors for AND, OR, and NOT?
  *   Facet checkboxes: Media Type — Sound or Moving Image, Contributing Organization, and State.
  *   Do you have any suggestions on better placement of search tips on the page? I think it would be good to include search tips somewhere. Maybe below the search box as in the other example I sent from University of South Carolina MIRS?

PR #490, opened 2015-07-01
Captions 444

mccalluc:
How exactly captions will work may change, but I think defining the interface at this point is safe. 
Though, if you think this is premature, I wouldn't argue that, either.

### DIFF
--- a/jetty/solr/blacklight-core/conf/schema.xml
+++ b/jetty/solr/blacklight-core/conf/schema.xml
@@ -506,8 +506,12 @@
           type="string" indexed="true" stored="true" required="true" />
    <field name="timestamp" 
           type="date" indexed="true" stored="true" default="NOW" multiValued="false"/>
+   
    <field name="xml" 
           type="string" indexed="false" stored="true" multiValued="false"/>
+   <field name="captions"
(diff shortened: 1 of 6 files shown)


jetty/solr/blacklight-core/conf/schema.xml
```diff
@@ -506,8 +506,12 @@
           type="string" indexed="true" stored="true" required="true" />
    <field name="timestamp" 
           type="date" indexed="true" stored="true" default="NOW" multiValued="false"/>
+   
    <field name="xml" 
           type="string" indexed="false" stored="true" multiValued="false"/>
+   <field name="captions"
+          type="string" indexed="false" stored="true" multiValued="false"/>
+
```
mccalluc: Actually, we decided *not* to index captions, I believe. Remove this line.

afred:
@mccalluc I'm going to rebase this, unless you already started to.

mccalluc:
Feel free, and then merge... I think we know that there will be tweaks when we actually have data, but this works as a framework, at least. (Reminder that it's putting the caption logo on everytime, regardless of whether a transcript is actually available. Ok with me if we tolerate this, but it's not perfect.)

PR #492, opened 2015-07-01
Configure exhibit in Markdown

mccalluc:
Exhibit configuration would be getting too complicated to expect it to be maintained in yaml by non-devs. So we use markdown and parse out the pieces we need. Other advantage here is that they can use github's built in preview to get a sense for their changes.


mccalluc:
@afred: I don't think this one is fine to merge, unless you see problems.

mccalluc:
@afred: I'll try rebasing this off the new master. Might be a new PR...

mccalluc:
Close in favor of #525

PR #496, opened 2015-07-06
Removes all Capistrano deployment stuff.

afred:
Deployment code moved to private repo `aapb_deployment`,
available within WGBH network.


afred:
@mccalluc I re-changed my mind on doing this. The main reason is that it makes it easier to upload secret config files from a private repo by assuming that they are sitting right alongside the deployment code.

mccalluc:
Sounds good. Merging.

PR #515, opened 2015-07-09
Remove duplicate values. Fix #482

### LINKED ISSUE #482: Discuss: Metadata redundancy in descriptions
For a lot of WGBH digital assets, there are two of the same descriptions appearing in the web application. Here is an example: http://americanarchive.org/catalog/cpb-aacip_15-rj48p5vm8b.

This is a problem with the metadata. In the AMS, I see that one description type = "Summary3" and another description type = "Episode Description."

Does this need to be cleaned up in the AMS before it can appear cleaned up in the web application? 
@sroosa  is working on creating descriptions for "Series," so this would be a higher level description than an episode description. But it is preferred that each unique description only appear once in the metadata record.

mccalluc:
@caseyedavis12 / @sroosa : Have we seen this problem in any field beyond title, description, and instantiation identifier? Not a problem to make it broader... but it shouldn't be applied to every single element: ie, it's not a problem if multiple instantiationDurations have the same value.

### DIFF
--- a/scripts/lib/cleaner.rb
+++ b/scripts/lib/cleaner.rb
@@ -223,6 +223,20 @@ class Cleaner # rubocop:disable Metrics/ClassLength
       Cleaner.clean_language(node)
     }
 
+    # duplicate value removal
+    
+    seen_values = Set.new
+    ['/pbcoreTitle', '/pbcoreDescription', #
+        '/pbcoreInstantiation/instantiationIdentifier'].each { |name|
+      match(doc, name) { |node|
+        if seen_values.include?(node.text)
+          Cleaner.delete(node)
+        else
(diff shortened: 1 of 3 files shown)


sroosa:
@mccalluc I have seen definitely seen dupes in genre. I think we should do it a little broader, but I see your point about not wanting to do it across the board. 
Fields I suggest removing duplicates:
Title
Description
Genre
Subject (do we display this?)
Contributor (but not role)
Creator (but not role)
Publisher (but not role)
Asset Type (this should not happen, but hey just in case)
Rights Summary

mccalluc:
(Actually, just going to do genre and rights summary after talking with Sadie.)

PR #516, opened 2015-07-09
Makes download dir path consistent.

afred:


afred:
@mccalluc ping

mccalluc:
actually, is this the line you're making it consistent with?
> MountValidator.validate_mount(Dir.pwd, 'downloads') unless is_same_mount

That's actually just a message: It might be more clear to change that to 'pbcore downloads' to make it obvious that it's not a fragment. What I'm doing with the direcotory creation is also confusing... I'll add to the PR, hold on.

afred:
oh.. right on.. I just know that the symlinked dir was still there, but it was called `downloads`, and the big fat dir that gobbled up all the disk was called `download`

mccalluc:
@afred : Maybe this is easier to read?

Issue #485, opened 2015-06-30
Search by Map


mccalluc:
Just an image which has
- hover style for states and individual station push pins
- clicking does a facet constrained search, either to state or individual station.

Question:
- Where does it live? Is front page too long already?

TODO:
- We need a state facet. The data is clean; we just need to pull it out.

caseyedavis12:
@mccalluc: If possible, we would like for the "click" on the station push pin to take the user to that station's page.

We would like for the map to go above the list of stations on the Participating Organizations page. Above the map, could we have a link that says "Explore list of stations" or "Explore map"?

mccalluc:
Map is deployed; Folks are happy with it.

PR #518, opened 2015-07-10
Add map to organizations page.

mccalluc:
Towards #485.

- Is this good enough? We could go the whole map-api/tiling-imagery approach, but I never see one of those used for simple navigation without thinking it overwhelms the rest of the page.
- Also, with map-api/tiling, you're stuck with a Mercator projection, which is just not a good way of showing lower 48+Alaska+Hawaii+Guam.
- If SVG is used as the src of an img, <a> tags and hover style do not work, so, include the whole thing.
- Hover style is a bit wonky, because the states are actually overlapping in a strict order: a wider border on a lower one can simply be obscured by a higher one. Could bring on more JS, or maybe change the color instead.
- Treatment for states / territories without content? Maybe no visual difference, but just don't make them a link? Giving them a special color just draws attention.
- Showing Guam without PR seems a little weird, but is showing all those islands weirder?
- Push-pins: I started on that, but then realized there's just not that much space here.

screen shot 2015-07-10 at 3 32 12 pm


mccalluc:
TODO: need to remove wrapping <a> from everything that doesn't deserve it.
(At some point we might think about generating the SVG in the code, but for now, I think we're better off keeping this as a static file, even if it's loaded in a weird way.)

mccalluc:
@caseyedavis12 : Thinking about it, one thing that's missing right now is a sense of the geographic coverage of the AAPB: I think a good solution would be little dots (the same size as the circle of DC) for each station, and they would glow when you hover on the state, along with the state itself. I don't think it would be good to make the stations dots clickable, actually: A lot of points are going to fall nearly on top of each other, and suggesting that users move the mouse pixel-by-pixel will just be frustrating, when they'd have better luck just scanning the state list. --> It's just about showing the geographic range of the AAPB

ok? Or are just state outlines enough?

afred:
I think this is a fine approach so long as we never have to zoom, pan, or pinpoint anything other than the stations. If we foresee ever wanting to do map-based stuff of that nature, then using a maps-api is better.

Also, if our map "points" are circles that represent cities, _and_ we have multiple stations within a city, _and_ we want the user to be able to click the map and get to a station's page -- then we're going to have to get clever on how that works.

thoughts @caseyedavis12 @mccalluc

mccalluc:
I don't think we should support clicking on individual station push pins.

My thinking right now is that having a cloud of push pins that the user needs to hunt through is really an anti-pattern: It's only the ubiquity which makes us forget how poor a solution it is. None of our states have so many stations that the list for a state can't be taken in at a glance. I'm totally for dots on the map to suggest our coverage, but I don't think they should be navigation elements.

I don't think the normal Mercator project we'd get with a maps-api is the best way of presenting this: I'm glad we've got Guam, but I don't want it to be at the cost of making the east coast unusable. (Alternatively, you could have the API zoomed in to the lower 48, but then you lose the outliers.)

afred:
yeah, using a maps-api would make an "all at once" geographic view not very possible. But then again, you'd only use the maps-api for a larger degree of user interactivity, and the "all at once" wouldn't be as important.

i just want to get from @caseyedavis12 what she has in mind for this map. If there is additional user-interaction planned for the next 6 months (e.g. zoom, pan, pinpiont, etc), I would recommend at least considering a maps api.

If additional user-interaction is not planned, or is more than 6 months off, then lets stick with this SVG, which is great btw.

caseyedavis12:
@mccalluc and @afred: will have more on this tomorrow.

mccalluc:
@caseyedavis12 : I think you had said this was good, but we weren't clear on where it should go?
- Don't think it would hurt anything on the org page.
- Ok on homepage, too, as far as I care, but maybe get rid of the org carousel then? They would be redundant, and the page is getting long...

Also, appearance questions:
- Color scheme / line thickness / etc all good?
- Confirm that stations will be dots, but no special hover behavior or clickability?
- Do you want Guam on there? Do you want PR and Virgin Islands as well? I don't want to slight Guam, but I don't want it to be too cluttered either.

mccalluc:
Close in favor of #619

PR #520, opened 2015-07-13
Coverage

mccalluc:
Add simplecov, and improve test coverage.


app/models/vocab_map.rb
```diff
@@ -78,8 +78,4 @@ def map_reorder_nodes(nodes)
     }
     nodes.each { |attr| attr.element.parent.delete(attr.element) }
   end
-
-  def self.delete(node)
-    node.parent.elements.delete(node)
-  end
 end
```
afred: ^^ removing because unused?

scripts/lib/pb_core_ingester.rb
```diff
@@ -80,10 +80,6 @@ def ingest(opts)
     end
   end
 
-  def optimize
-    @solr.optimize
-  end
-
   def commit
```
afred: ^^ removing because unused?

afred:
rebasing. stand by...

scripts/lib/pb_core_ingester.rb
afred: ah.. yes, i should check the commit msgs before commenting.

afred:
@mccalluc I appear to have completely jacked up this PR by trying to do something relatively simple.

The `master` branch had drifted on this PR, and required a merge or rebase. I rebased and am guessing I messed it up.

If you still have your `coverage` branch locally, can you `git push -f` it here?

mccalluc:
@afred: erg: I may regret having done this:
```
 + 7a30560...a2aa9a5 coverage -> coverage (forced update)
 + a846d12...85a7baa master -> master (forced update)
```
Maybe I clobbered some of your merges to master?

If I had it to do over, I would have just made a new branch with the changes I had locally.

afred:
@mccalluc hmm... uh, let me check something. stand by...

afred:
@mccalluc Ah.. yeah. It looks like it may have undone some recent PR merges. Not too big of a deal (i don't think). We should be able to just re-open them and re-merge them. I hope.

I should have been more explicit and said `git push -f --dry-run` first.. or `git push -f origin coverage`. My local git is config'd to only push the branch you are on. It's one of those things I specify once, and then don't think about again until I get a new computer. Sorry about that.

I'll see about re-opening the PRs.

afred:
@mccalluc ok, so you can re-open PRs that have been closed _without_ being merged, but you _cannot_ re-open PRs that have been merged. For those, your options are to restore the branch the PR came from (if one exists) -- OR -- you can "revert" the PR, which is actually creating a new PR with the reversed-diff. We definitely do not want the latter.

I think what I'll do is restore the branches on the lost PRs, merge them with master locally, re-commit them and re-submit new PRs for them. There are only a few.

afred:
@mccalluc i think i found a piece that slipped through the git voodoo from earlier. I had to rebase this onto the new `master` branch because it contained commits from the old `master` branch that no longer exist. But after the rebase, a couple of tests failed. I fixed one of them but not sure about this one.

Error is caused by `Uncataloged: 1991-07-27` being in the expected value, but not the returned results. It's not clear whether the expected value is supposed to be correct, or if the search results are supposed to be correct. Also, not sure where the search results are coming from. Fixtures i presume, but which ones?

afred:
^^ I made the assumption that the expected value needed changing. Check out the last commit in this PR, and if it looks good please merge. The rest is yours.

PR #522, opened 2015-07-14
Replace links that are redirects.

mccalluc:
This quiets a once-a-week test, bu otherwise is not too useful. I'm ambivalent about this.


afred:
it appears that your actually fixing links in website content.. which would seem very useful. Is that not what's going on?

mccalluc:
So all of these are redirects; none are 404s: the test is just picky. In some cases I think the original url is actually "better" in some sense, but that's subjective. That's all.

afred:
ah. thanks for the clarification.

PR #525, opened 2015-07-16
New, clean branch bringing over all the changes from #492

mccalluc:


mccalluc:
@afred: if you could take a look at this, I would really appreciate it. I don't think there are any links to this from any other pages of the live site, so it's low risk, but if we keep it out longer there will be new merge problems. Thanks!

spec/support/validation_helper.rb
```diff
@@ -33,7 +33,13 @@ def expect_fuzzy_xml
...
     
     # try to parse as xml
-    REXML::Document.new(xhtml)
+    doc = REXML::Document.new(xhtml)
+    
+    title_node = REXML::XPath.match(doc, '/html/head/title').first
+    raise 'Page should have title' unless title_node and !title_node.text.empty?
+    if require_distinct_title && title_node.text == 'American Archive of Public Broadcasting'
+      raise 'Page title should be distinctive'
+    end
```
afred: Coding style suggestion: in general, using an `options` hash for optional parameters makes methods easier to use, and more readable when they are being called. E.g.
```ruby
# instead of...
def expect_fuzzy_xml(require_distinct_title=true) do
end

# you can do this
def expected_fuzzy_xml(options={})
  require_distinct_title = options.delete(:require_distinct_title)
```

afred:
Re: the data layer vs presentation layer for exhibits...

Years of MVC development forces me to be wary about coupling the data layer and presentation layer as tightly as they are in `app/models/exhibit.rb`, particularly [here](https://github.com/WGBH/AAPB2/blob/exhibits-try-again-484/app/models/exhibit.rb#L109-L147).

I'm worried that the tightness of the coupling here is going to put a limitation on how complex we allow the presentation layer to get.

Down the road, if we find that playing around with the markdown in new ways starts to break the model layer, we should consider a different approach.

afred:
Also... it would help me out a lot as a code reviewer if tests are described a little less ambiguously. E.g. a test described as "works" doesn't tell me anything, and it takes longer to figure out what it's supposed to be doing.

afred:
@mccalluc are we going to have real Exhibits right away? Or should we get rid of the sample ones (midwest, iowa, et al) and then handle the case of 0 exhibits, say, with a quiet redirect, or an actual msg saying "There are currently no exhibits". Thoughts?

PR #529, opened 2015-07-22
S3 urls

mccalluc:
@afred : look ok? Do we want a CNAME? Or a constant in the code?

### DIFF
--- a/app/views/exhibits/template.erb
+++ b/app/views/exhibits/template.erb
@@ -15,7 +15,7 @@
             <% list.each_slice(2) do |value,count| %>
               <li>
                 <span class="facet-label">
-                  <a href="/catalog?f[exhibits][]=<%= @exhibit.path %>&amps;f[<%= name %>][]=<%= value %>" class="facet_select">Moving Image</a></span>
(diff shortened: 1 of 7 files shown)


app/views/exhibits/template.erb
```diff
@@ -15,7 +15,7 @@
             <% list.each_slice(2) do |value,count| %>
               <li>
                 <span class="facet-label">
-                  <a href="/catalog?f[exhibits][]=<%= @exhibit.path %>&f[<%= name %>][]=<%= value %>" class="facet_select">Moving Image</a></span>
+                  <a href="/catalog?f[exhibits][]=<%= @exhibit.path %>&amps;f[<%= name %>][]=<%= value %>" class="facet_select">Moving Image</a></span>
```
afred: `&amp` instead of `&amps` ?

afred:
Yeah, a constant or config bar would be appropriate.

```ruby
module AWS
  S3_BASE = "http://americanarchive.org.s3.amazonaws.com/"
end
```

i've never really settled on a preferred pattern for app/module config. There are a lot of them out there. I'm open to suggestion.

mccalluc:
@afred: When the test is done, I think this is good to merge?

PR #531, opened 2015-07-22
Relative, floating links on the homepage

Hammicakes:
update to homepage adding relative floating links for navigation ON the homepage.


afred:
This PR contains a bunch of commits that aren't part of the feature being described. I think this is usually due to a bunch of merging from other (un-merged) branches prior to making the PR.

If this feature requires code from PRs that haven't been merged yet, it's better to wait until they have been merged, even though they my be blocking.

If the feature does not require the work from unmerged PRs, it's better to go back to a clean `master` branch, create a new feature branch, and do the work there. That way the PR will be much easier to review.

mccalluc:
@afred: I'm confused about the state of this... Does something else need to be done here?

afred:
@mccalluc the PR is too big to review -- but only because it's including a bunch of stuff from other commits.

The PR needs to get down to a size where it can be reviewed. This can be accomplished with a rebase probably.

PR #538, opened 2015-07-24
537 remove jetty and move solr config

afred:
@mccalluc there are now two deployment tasks that will:
  1. download a fresh copy of jetty
  1. copy `solr_conf/solrconfig.xml` and `solr_conf/schema.xml` into the jetty instance


afred:
And I think there will have to be a little tweaking on Travis to ensure jetty gets downloaded. I have code for this I can pull from hydradam. Stand by..

mccalluc:
One thought: Travis has its dependency caching now, so I wonder if that would influence how to approach this? Maybe somehow the jars are checked into a gem, rather than being re-downloaded with each build?

afred:
Yeah maybe. If the builds prove to be too slow, then we can look into it.

config/jetty.yml
```diff
@@ -1,11 +1,11 @@
...
 development:
-  startup_wait: 15
-  jetty_port: 8983
+  <<: *default
 test:
-  startup_wait: 60
-  jetty_port: <%= ENV['TEST_JETTY_PORT'] || 8983 %>
-  <%# TODO: Get separate dev and test cores, and change this back to 8888. bug #1 %>
-  <%= ENV['TEST_JETTY_PATH'] ? "jetty_home: " + ENV['TEST_JETTY_PATH'] : '' %>
-production:
```
mccalluc: @afred: Do we still want `production` to get the default, or is it supposed not to be on this list?

lib/tasks/ci.rake
```diff
@@ -0,0 +1,38 @@
...
+    # The :startup_wait value is the number of seconds Jettywrapper will wait
+    # while checking to see if jetty started. A high value helps ensure Travis
+    # builds don't time.
+    :startup_wait=> 180
+  })
+  
+  # make sure we've got the thing downloaded
+  puts 'Checking to see if jetty download was successful...'
+  sleep(1) until Dir.exists?(jetty_params['jetty_home'])
+  puts 'Jetty download detected.'
```
mccalluc: Some asynchronous process started the download? Depending on how it's done, I could easily imagine there being a directory without all the files in place yet...

lib/tasks/jetty.rake
```diff
@@ -0,0 +1,23 @@
+require 'pry'
+require 'rails_stub'
```
mccalluc: Are either of these necessary? I guess pry is a convenience, but I was thinking of rails_stub as something that just comes into play when a test involves a class which has a superficial dependence on Rails... Not clear that it's needed here?

afred: just bumped into another issue with this rake task though. Fixing it.. stand by.

lib/tasks/jetty.rake
```diff
@@ -0,0 +1,23 @@
...
+    # Jettywrapper.load_config is convenience method for loading
+    # config/jetty.yml
+    params = Jettywrapper.load_config
+    
+    config_filenames = ['solrconfig.xml', 'schema.xml']
+
+    config_filenames.each do |filename|
+      cmd = "cp #{File.join(Rails.root, 'solr_conf', filename)} #{params['jetty_home']}/solr/#{params['solr_core']}/conf/#{filename}"
+      puts cmd # print the command to stdout
+      output = `#{cmd} 2>&1` # run the command, capturing stdout and stderr
```
mccalluc: Backticks make me nervous... I think FileUtils.cp_r would work, and we wouldn't even need the loop.

afred: right on. Stand by.

lib/tasks/ci.rake
afred: Travis failed with a "there is no jetty" error, and then it passed after putting in the sleep step. Because it downloads a zip file and then extracts, I'm waving my hands and saying that looking for the directory is sufficient -- even though unzipping takes longer than zero seconds.

Maybe this isn't needed, and the missing jetty was due to some other false negative. I'll try again without it. Maybe it would be better to let if fail anyway, instead of hanging forever waiting for a download that may never come due to some other reason.

config/jetty.yml
afred: @mccalluc production isn't on the list.

mccalluc:
I'll try using this locally, and put some instructions in the readme, unless you have plans.

afred:
Just limit the instructions to those applicable on local machine / development environment. I'm updating the docs in the deployment repo for production/demo environments.

mccalluc:
```
$ bundle exec rails c
/Users/chuck_mccallum/.rvm/gems/ruby-2.0.0-p353@global/gems/bundler-1.6.1/lib/bundler/runtime.rb:34:in `block in setup': You have already activated spring 1.3.6, but your Gemfile requires spring 1.3.3. Prepending `bundle exec` to your command may solve this. (Gem::LoadError)
```
Not a problem with this pull, per se, but I want to know I can start up rails before merging.

afred:
your guess is as good as mine on that one

afred:
[Spring](https://github.com/rails/spring)... this look like one of those things that is supposed to magically make things work better, until it breaks and prevents anything from working at all.

mccalluc:
@afred: review my tweaks, and merge if it still looks good to you? (After the merge, thinking that the rake tasks should be pulled into a gem and applied to stock-sales...?)

afred:
:+1: for a gem for the rake tasks. Call it `jetty-helpers` or something like that. When the time comes... [How to load rake tasks from a ruby gem](http://stackoverflow.com/questions/28010171/load-rake-tasks-from-ruby-gem)

config/jetty.yml
mccalluc: @afred: I'm still a little confused about this part. Production just doesn't use this at all? Grab me after lunch?

PR #545, opened 2015-07-28
Advanced search again 478

mccalluc:
Looking at this, I should add some tests, but wanted to confirm that this is what folks imagined. @caseyedavis12 / @sroosa?

screen shot 2015-07-28 at 4 35 04 pm

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,4 +3,7 @@ sudo: false
 cache: bundler
 rvm:
   - 2.0.0
+before_install:
+  - 'export DISPLAY=:99.0'
+  - 'sh -e /etc/init.d/xvfb start'
 script: 'bundle exec rake ci'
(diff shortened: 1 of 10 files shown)


caseyedavis12:
This is awesome.


The only thing I see that maybe @hammicakes can work on is that now the search bar in the header is really close to the logo.

mccalluc:
Thanks! hmm... Way back when we reduced the vertical spacing above the logo to fit more on the page. options:
- undo that tweak.
- make the search box not as wide.
- Something shorter than "Advanced Search..."
- Do nothing: This was a particularly small window, just for the screenshot.

Glad to defer to someone else, but I don't think that needs to block merge.

caseyedavis12:
@mccalluc: Ok, I didn't realize that the screenshot was a small window. I vote for:

  *   Do nothing: This was a particularly small window, just for the screenshot.

mccalluc:
Thinking about it overnight, the UI is fine, but the query construction would be better done on the server: 
- No dependence on JS.
- Less testing overhead.
(Closing, but will keep the branch until I'm sure the other way will work.)

PR #547, opened 2015-07-29
View all results from state

### LINKED ISSUE #543: Add see-all-per-state link at the top of each section on the orgs page.
This is related to the map, but can be done independently.

mccalluc:
Links from org page.
New facet, but not included on sidebar? ... Is that right?
Fix #543.
screen shot 2015-07-29 at 2 15 50 pm


caseyedavis12:
@mccalluc It would be good to also include the states list in the facet sidebar if possible

afred:
@mccalluc want to add state facet to sidebar in this PR or another?

PR #552, opened 2015-07-30
Add availability constraint to genre browse

### LINKED ISSUE #528: add an availability constraint on front page browse list

mccalluc:
fix #528: UI might change with input from Mike H, but something like this will still be in place.


afred:
@mccalluc this just adjusts the default results for faceted searches, right?

mccalluc:
@afred: Yeah, it's just making the homepage genre and topic links into narrower searches, as requested.

PR #555, opened 2015-07-30
Exhibit index page

### LINKED ISSUE #517: Clarify design and implement exhibits index page
@caseyedavis12 will clarify what the exhibits list page should look like.

Things to consider:
- Do all exhibits get listed? In an indented hierarchy? Or just the top level ones?
- Does any info about the contained items appear? (count?)
- Can we use the same summary for each that would occur on a constrained search page? (I hope so.)
- What order?
- Do we use the same sidebar as the "about" pages? Something different? Nothing?
- Does anything else go on this page? Header? sub headers?

I have some deep trepidation about just making it look like a search results page, because people will expect it to work like a search results page. On the other hand, if that's what we want, it wouldn't be impossible to make records from the exhibits, and I would much prefer that to getting successive requests to make it work more and more like a search results page. What you would lose then is the chance to treat the hierarchy of exhibits specially, or anything else that is particular to them.

mccalluc:
Fixes #517.

@caseyedavis12 : Looks like this: Let me know if there are problems.
screen shot 2015-07-30 at 5 56 31 pm

### DIFF
--- a/app/controllers/exhibits_controller.rb
+++ b/app/controllers/exhibits_controller.rb
@@ -2,7 +2,7 @@ require_relative '../models/exhibit'
 
 class ExhibitsController < OverrideController
   def index
-    @exhibits = Exhibit.all
+    @top_level_exhibits = Exhibit.all_top_level
     @page_title = 'Exhibits'
   end
   
--- a/app/models/exhibit.rb
+++ b/app/models/exhibit.rb
@@ -12,6 +12,11 @@ class Exhibit < Cmless
   
   attr_reader :head_html
   
+  def self.all_top_level
+    @all_top_level ||=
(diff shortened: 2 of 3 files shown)


caseyedavis12:
Thanks, @mccalluc!


Did you increase the size of the thumbnails on the Exhibits page (larger than the standard Grid View), or do we need @hammicakes need to do this?


Having two thumbnails per row on Exhibits would be good. It would increase the size of the signature image for the Exhibit and also make the Exhibits page looks a little different than the standard Grid View search results.


[screen shot 2015-07-30 at 5 56 31 pm]<https://cloud.githubusercontent.com/assets/730388/8996306/fa7aa006-36e4-11e5-94d1-689b56968d3d.png>

PR #561, opened 2015-07-31
floating navigation on homepage

Hammicakes:
added a floating navigation to relative links on the homepage


app/assets/stylesheets/normalize.css
```diff
@@ -717,7 +717,9 @@ table {
 
     /* Homepage Welcome Header */
     #welcome h1 {
-      font-size:1.8em;
+      font-size:1.4em;
+      font-weight:;
```
mccalluc: @Hammicakes / @hamiltopia (which handle do you prefer?): Do you get notices for the test failures? This looks like where it's coming from... Since there are also merge conflicts, I'll handle it. Thanks!

Hammicakes: Not sure how @hamiltopia is in here.  @hammicakes should be the handle.  I am looking into it.

Thanks.

mccalluc:
@Hammicakes : a few other things
- I'm not sure what's easiest for you, but if sorting out the merge is difficult, you might try updating master and making a fresh branch. The only change I see on homepage is the change the addition of an advanced search link, and the removal of the field selector on search.
- New CSS is interfering with hovers on the galleries: instead of zooming inside the div, the div is getting bigger. 
- Also, the left and right scroll arrows on the galleries are getting a background on hover that wasn't there before.

PR #562, opened 2015-07-31
Attempt to clarify access facet.

mccalluc:
@caseyedavis12 / @sroosa : I think there's a consensus that the access facet needs more explanation to the user. I really don't think that having stand alone documentation will be useful: it needs to be at the point where people are confused. Here's an attempt to provide that:
screen shot 2015-07-31 at 3 48 26 pm
- Is the idea reasonable?
- Is the style ok? (Maybe just on hover?)
- Is the text ok? Maybe a link to other docs if more needs to be said?


mccalluc:
2 more thoughts:
- Order: ORR at top (default/most restricted), then on-site, then all.
- Selection: Make it radio-buttony. Instead of the BL's filter add-remove idiom, One choice has a check at all times, and clicking to another moves the check.

Looks like:
```
✓ Online Reading Room  2
       ORR blurb goes here...
   On site only        4
       On site blurb here...
   All                 8
       All blurb...
```

caseyedavis12:
@mccalluc:


I really like where this is going.

I'd like to get folks' feedback on a few things -- @afred @sroosa

*Should we keep this facet always expanded?
*"Access" is not always understood by the user. Can we brainstorm a better term to use for this facet?

Here are edited blurbs below. Any suggested further edits?

1. Online Reading Room -- Maybe change to "View/Listen Online"
"The default: shows only records for video and audio that can be viewed or listened to in the Online Reading Room."

2. On site only - does this facet include all ORR and all on-site only (all digitized)? I think it is best to include On-Site and Online in this facet. If so, I would change this to "Online & On-site Only"
"This shows records for all video and audio that has been digitized, including content available online and content that is restricted to viewing/listening at WGBH and the Library of Congress only"

3. "All"
"This shows all approximately 2.5 million records, including for all digitized content (online and on-site only), and for content that has not yet even been digitized."

afred:
:+1:

caseyedavis12:
@afred @mccalluc @sroosa 

I like all of Drew's wording edits, except I question "On-site access only." If this filter includes stuff that is available in the ORR and on-location only, then we should clarify that this is a larger umbrella filter for all digitized content (online and on-site).

sroosa:
@caseyedavis12 @afred @mccalluc: Couldn't we just leave out the "only"? and also maybe we should switch the wording to On Location, since that's what we use throughout the rest of the website.

afred:
:+1: for "on location" instead of "on-site"

maybe it's just me, but I think that people will most likely want to avoid returning results for which they have to be "on location". As such, I think there is value for a higher number of people in keeping it "On Location Only".... which subtly says "warning, you won't be able to watch/listen to these in your web browser".

i would think that the only time people would be interested in seeing the "On location" records, is if there is a chance that they'll actually be able to get to WGBH or LoC to access them. And on average, I'm guessing that chance will be small... less than 1%.

But that's just.. like... my opinion, man.

caseyedavis12:
@afred: I was trying to clarify the 2nd tier of access. We should definitely have first tier of access be Online records only. I was questioning whether the second tier of access should include Online and On Location (all digitized assets) or just On Location assets. My vote is for all digitized assets, including online and on location only.

afred:
@caseyedavis12 yeah, I understand that. What i was trying to get across was that I think it is less useful to the average user to combine "Available Online" and "Available On Location" into the same faceted result set.

but... for that matter... why can't we "OR" these together like other facets? Wouldn't that give the most control with the least amount of confusion?

caseyedavis12:
@afred @mccalluc @sroosa 
Ok, let's try this again. Since we are keeping the facet expanded on default, maybe keeping "Access" as the facet term is fine.

"Available Online":
"Audio/Video can be played in the Online Reading Room."

"Online & On-Location"
"Broaden your search to include audio/video that can only be played at WGBH in Boston and the Library of Congress in Boston."

"All Records"
"This includes records for audio and video that have not yet been digitized."

mccalluc:
(Hold on, more UI work needed.)

mccalluc:
Here's what it looks like now:
screen shot 2015-08-03 at 5 33 41 pm

The most substantive addition has been to add this facet to the search if it wasn't already present, by doing a redirect. I feel like, in some sense, it would be better if the default state, searching public, digitized content, did not add clutter to the screen, but I think that would be more involved than this.

Also got rid of references to search_field, which isn't being used any longer.

mccalluc:
Feeling more persuaded to @afred's way of thinking. Here's what it looks like:
screen shot 2015-08-03 at 6 04 21 pm

(Let's not have any more conversation on this thread, and instead converge on something good-enough at the meeting.)

mccalluc:
Note: Adding potential query constraints is low cost, and we could add them to the UI later. For example, do we add a filter for just the private ones?

PR #563, opened 2015-08-03
adds new images to gallery

sroosa:
replaces old images, with images for records that will be in the ORR

### DIFF
--- a/config/featured/home-featured.yml
+++ b/config/featured/home-featured.yml
@@ -2,90 +2,131 @@
 # If the image is sitting in the expected place on the media server,
 # "thumbnail_url" can be omitted.
 
-- id: cpb-aacip_153-97kps2vw
-  org_name: Oregon Public Broadcasting
-  name: "In Search of the Oregon Trail"
-
-- id: cpb-aacip_254-1615dwj5
-  org_name: WPBS
-  name: "Rod and Reel Streamside: Deadly Techniques"
-
-- id: cpb-aacip_375-93gxdckz
-  org_name: Radio Bilingue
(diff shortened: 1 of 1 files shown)


afred:
@sroosa all these images are the same?.. i.e. `thumbs/audio-digitized.jpg`?

sroosa:
@afred I put the audio ones at the top and the video ones at the bottom, so that it's easier for me to change them out if I need to. The audio records all have the same image, but the video records each have a unique thumbnail.

mccalluc:
hmm... but is that necessary? I'll check out this branch and see if it still works without those. Hold on...

afred:
@mccalluc @sroosa I guess i'm unclear on why the videos don't have a value for `thumbnail_url`, while the audio does, but they are all the same. Clarification?

afred:
@mccalluc just clarified verbally. Thanks!

mccalluc:
(Talked in person. Waiting to see if Sadie can put the images on S3...)

mccalluc:
@afred : do you have any experience with s3 permissions? Somehow the "casey_davis" user was granted privs; I'd like to do the same for Sadie: I've tried both her user name and ARN... Can you think of something else to try? I've found threads like this that look unhappy: https://forums.aws.amazon.com/thread.jspa?threadID=70876#281141

afred:
@mccalluc i don't. Reading on it now.

just to be clear... we want to allow multiple people to push to a single bucket. Is that the use case?

sroosa:
@mccalluc I just added all of the new images to S3.

PR #565, opened 2015-08-03
Adds conditional around :ci rake task

### LINKED ISSUE #564: Running `rake` fails on production
**Reproduce:** On newly deployed code, try running...

  ```
  RAILS_ENV=production bundle exec rake -T
  ```

... and you get the following stack trace...

**Stacktrace:**
```
rake aborted!
LoadError: cannot load such file -- rspec/core/rake_task
/srv/www/aapb/shared/bundle/ruby/2.0.0/gems/activesupport-4.1.2/lib/active_support/dependencies.rb:247:in `require'
/srv/www/aapb/shared/bundle/ruby/2.0.0/gems/activesupport-4.1.2/lib/active_support/dependencies.rb:247:in `block in require'
/srv/www/aapb/shared/bundle/ruby/2.0.0/gems/activesupport-4.1.2/lib/active_support/dependencies.rb:232:in `load_dependency'
/srv/www/aapb/shared/bundle/ruby/2.0.0/gems/activesupport-4.1.2/lib/active_support/dependencies.rb:247:in `require'
/srv/www/aapb/releases/20150803184115/lib/tasks/ci.rake:1:in `<top (required)>'
...
```

Running `rake` will load all your rake tasks. In this case, it's loading the task `ci`, which is for Travis CI server, which requires `rspec/core/rake_task`... but `rspec` is only included when in the `test` environment (and rightfully so). So we need to avoid loading the `ci` task when in production.

**Done when**
Running `RAILS_ENV=production bundle exec rake -T` does not error.

afred:
Fixes #564

### DIFF
--- a/lib/tasks/ci.rake
+++ b/lib/tasks/ci.rake
@@ -1,33 +1,37 @@
-require 'rspec/core/rake_task'
+# Do not build this rake task when in production environment.
+if Rails && !Rails.env.production?
 
-desc "Run tests as if on CI server"
-task :ci do
+  require 'rspec/core/rake_task'
 
-  require 'jettywrapper'
+  desc "Run tests as if on CI server"
+  task :ci do
 
-  # Set the version of hydra-jetty we want, and download a clean copy of it.
-  Jettywrapper.hydra_jetty_version = 'v8.4.0'
-  Jettywrapper.clean
+    require 'jettywrapper'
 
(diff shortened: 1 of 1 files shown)


afred:
@mccalluc I tested this manually on the demo server, and seems to work. Tried a few different ways to test it in rspec, but couldn't get them to work.

Issue #569, opened 2015-08-04
`titles_sort` no longer needed, if `title` is concatenation


mccalluc:
With #508, the `title` will be a concatenation of all titles, from least to most granular. When we are convinced that that is stable, the `titles_sort` method can be removed. I will put a TODO in the code, but I don't want to jump there right now.

PR #570, opened 2015-08-04
Make the official title a semicolon delimited list.

### LINKED ISSUE #508: Show all titles concatenated in search results
- On search results, show all the titles on a single line, joined with commas.
- Since it's redundant, don't show the titles on search results broken out by type.
- No change to the details page.

mccalluc:
Fixes #508.

(Some misgivings about this, but it might be good enough. If we stick with this approach, there will be some follow-up to clean-up: #569.)


mccalluc:
This is putting "Uncataloged" titles first, which is not what we want. Talked over a fix with Sadie: hold on...

PR #573, opened 2015-08-04
TOS click through take 2

mccalluc:
Rebase of tos-click-through: Didn't want to do it in place in case something got confused.


mccalluc:
@afred : #575 has the full picture of the rights we want to enforce. Does it seem appropriate to you to use CanCan here, or does that not seem like the best approach?

mccalluc:
Close in favor of #603

PR #581, opened 2015-08-06
Footer updates

Hammicakes:
-Added social links to footer
-better formatting of HTML
-cleaned up some CSS styles
-update to aapb-site-sprite.png


mccalluc:
@Hammicakes : purely for information's sake, but what dictates the height of the sprite image? (aapb-site-sprite.png) Would something break if it were shorter? Or is this particular height a good, conventional default? Or something else?

(I'll merge this locally, and hopefully that will work.)

mccalluc:
@Hammicakes : The HTML has a social links section, but because of the CSS it is not being shown... Should it be visible?
@caseyedavis12 : Could you confirm that these are the footer links you want? I think you want NDSR abbreviated here, rather than spelled out? Do you care whether it opens in a new tab or not? (I'm ambivalent, but there was a merge conflict.) You do want the social links to show here?

(Rendered at different widths just for reference: I don't see any problems with the responsiveness.)
screen shot 2015-08-10 at 5 48 49 pm

caseyedavis12:
@mccalluc: For now, these are the footer links that we need. By October, we will add one or two more pages that will need to be added to the footer. If possible, please have NDSR open in a new tab, since it takes the user to a different site (same for the link in the header). Yes, we would like for the social links that Mike Hamilton worked on to show here.


<https://cloud.githubusercontent.com/assets/730388/9185256/4d313908-3f88-11e5-8b27-8a814adfbabc.png>

mccalluc:
@Hammicakes : We'll hold this until the social media links are showing... I'm not sure what your plans were for them.

afred:
@Hammicakes hey, did you get the social media stuff to shop up in the footer (see @mccalluc input above).

mccalluc:
Master has diverged far enough from this that I couldn't work out the merge. Instead, just pulled out the parts necessary for the links and submitted that as #672.

PR #582, opened 2015-08-06
Homepage Navigation

Hammicakes:
Adding a relative, sticky nav for the sections on the homepage

### DIFF
--- a/app/assets/stylesheets/homepage.css
+++ b/app/assets/stylesheets/homepage.css
@@ -32,75 +32,65 @@
 		width:98%;
 	}
 
-/* Homepage Relative Nav */
-/*This is on hold for the time being*/
-.sticky-well {
-	background-color: rgba(255,255,255,.25);
-	border:0;
-	border-radius:0;
-	box-shadow:none;
-	height:50px;
-	margin:2em 0;
+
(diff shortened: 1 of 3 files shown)


afred:
@Hammicakes, can you please attach a screen grab for UI changes like this one? It's hard to see what's happening from the css alone.... or... at least it is for a non-designers like myself :smile:

afred:
^^ and by "attach", I mean paste it into the comments for the PR. Thanks!

mccalluc:
Looks good to me. Observations:
- We're getting a row of pixels for the next block row on some of the blocks in the header: just a tiny bit of orange at the bottom of the purple. Maybe if the image was resized at some point the border got dithered? Hardly noticeable, but wanted to mention.
- I think this is by design, but the "white" space above the section titles has increased with this: I think it's necessary so that the hover banner looks good. Only mentioning because at first I thought this was only changing the header, and that's not true.
- @caseyedavis12 : For the first two, the section title matches the link title, but the last one is "Telling Our Story" vs. "Every Picture Tells a Story". That's ok with you? (In my mind it's unclear whether "our" is referring to AAPB, or American more broadly.)
screen shot 2015-08-10 at 6 15 57 pm
None of which should block merging.

PR #583, opened 2015-08-06
Added Curated Exhibitions

Hammicakes:
-Added new section to homepage
-Added new carousel, copied from the featured carousel on homepage
-Added section to Exhibit page
-Added a link to be displayed on a record that is an item included in
an exhibition.  the example is on the Exhibition page: Midwest!
-Added staff profile photos: on staff page and exhibition page


mccalluc:
We need to identify the useful parts of this and make separate PRs for them. 
- I think it would be a big headache to try to maintain multiple branches in development, so we can't put the gallery on the homepage until there are exhibits ready to go. (We already have a commented out gallery: We just need to be about to un-comment that, when the time comes.) 
- similarly with the link from item to exhibit.
- lorem ipsum shouldn't be checked into the ERB: if there's a slot we can't fill from the the MD, that needs to be done first.

I'm going to make some smaller PRs and issues from this before I go.

PR #592, opened 2015-08-10
Updates bit about OpsWorks in README.

afred:


mccalluc:
Test failure is unrelated. I've filed #594.

PR #599, opened 2015-08-10
Script was failing with bad constant

mccalluc:
... not sure why we did not see problems earlier? Maybe running on branch without this refactor?


mccalluc:
fixes #600

PR #602, opened 2015-08-11
Staff-pics

sroosa:
I've added all of the ones I have so far. @mccalluc: Can we merge this now, and I can add the rest when they come in later?

### DIFF
--- a/app/views/override/about-the-american-archive/staff.md
+++ b/app/views/override/about-the-american-archive/staff.md
@@ -2,7 +2,7 @@
 
 ![banner](/page-banners/banner5.jpg)
 
-![karen](https://s3.amazonaws.com/americanarchive.org/staff/Staff_Cariani.jpg)
+<img class="img-circle" src="https://s3.amazonaws.com/americanarchive.org/staff/Staff_Cariani.jpg"/><br/>
 **Karen Cariani, Project Director**<br/>
 **WGBH**<br/>
 Karen has been director of the WGBH Media Library since its inception in 1990. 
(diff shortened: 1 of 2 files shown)


app/views/override/about-the-american-archive/staff.md
```diff
@@ -76,27 +78,27 @@ the outdoors.
...
 **WGBH**<br/>
 Sadie focuses her attention on metadata management and cataloging for the AAPB. 
 She has worked at WGBH since 2012, engaged in digitization and digital library 
 projects including Boston TV News Digital Library and WGBH's Open Vault. She 
 co-chairs the Digital Commonwealth Outreach Committee and the AMIA News, 
-Documentary and Television Committee. Even though Sadie spends the majority of 
-her work days cataloging at a television station, she also spends a significant 
-portion of her free time participating in crowdsourced cataloging initiatives, 
-while watching TV at home. Sadie earned her MLIS from the University of 
+Documentary and Television Committee. In her free time, she enjoys knitting baby booties, even though she knows very few people with babies, and rearranging furniture in her apartment, even though there is really only one place each things fits. Sadie earned her MLIS from the University of
```
mccalluc: s/things/thing/

Also, in general, it is good to put hard returns to wrap the lines at 80 columns or so: it won't change how it renders, but it will make it much easier to understand the diffs for future edits.

mccalluc:
@sroosa: There aren't actually any images in this commit -- Should there be? ... but before you add them here, could you just create a new directory and add them on S3? Then, if it's working, you should see your images when you preview it here at github.

sroosa:
@mccalluc just added all of the urls to the images which were already on S3. I can see them now, and verify they're working. Thanks!

mccalluc:
(Thought about just putting the CSS in the MD, or trying to make a CSS pattern that wouldn't require a new class, but both of those just felt even more awkward.)

PR #603, opened 2015-08-11
Access control 575

mccalluc:
In progress. Currently no change in behavior. Pull can wait until it incorporates cancan.


mccalluc:
@afred: please review and merge if it seems good. There is no real change in behavior, but it is based on cancan now, and when we do have TOS, it will be easy to adjust the Abilities to reflect the new rules.

mccalluc:
This is ready to go. When deployed:
- Outside users experience no change: they still have no access.
- WGBH users get the click-thru.
- At the end of the month, we can uncomment the right lines in Ability.

afred:
@mccalluc changing a couple of things. Stand by for update.

PR #608, opened 2015-08-12
Populate exhibit

mccalluc:
@mccalluc should do a sanity check on this, and then delete the mock exhibits he had made now that there is real data!


mccalluc:
@afred : This started off with just markdown, but there are enough real changes now that it merits a review.
@sroosa : Note that the tests now depend on some characteristics of your exhibit. This shouldn't get in your way, but as you make changes going forward, it might be better to make them in a branch before merging. (If it's just MD, feel free to self merge if tests pass.) Thanks!

PR #617, opened 2015-08-12
Adds specs for User#usa?.

afred:
Refactors IP range tests to use RSpec idioms.
(I know this is failing, but wanted to push to save changes. Will revisit later)


spec/models/user_spec.rb
```diff
@@ -1,36 +1,60 @@
...
-      expect_unauthorized(below)
-      expect_authorized(bottom)
-      expect_authorized(top)
-      expect_unauthorized(above)
-    end
-    it 'allows appropriate WGBH access' do
-      expect_ip_range('198.147.174.255', '198.147.175.0', '198.147.175.255', '198.147.176.0')
+      # Lists of all on-site IPs from WGBH and LOC
+      let(:wgbh_ips) { IPAddr.new('198.147.175.0/24').to_range.to_a.map(&:to_s) }
+      let(:loc_ips) { IPAddr.new('140.147.0.0/16').to_range.to_a.map(&:to_s) }  # NOTE: this is > 65K IPs
```
mccalluc: `IPAddr.new('140.147.0.0/16').to_range.minmax.map(&:to_s)` doesn't test the interior, but feels good enough to me.

mccalluc:
Thanks for doing this: looks like a good direction.

afred:
@mccalluc For some reason, looking up IPs for supposedly non-US websites always seems to return US ip addresses. Not sure why. Do you know?

for the non-US ip in the test, I just grabbed on from a [list on the web](http://www.nirsoft.net/countryip/).

afred:
@mccalluc review and merge if it looks ok

mccalluc:
@afred: A little confused by your comment: geo_i_p_country_spec still passes, as far as I know.

But looks good. Merging.

afred:
@mccalluc hmm.. i guess i was just trying out a bunch of sites that I thought would be non-US, but was getting US IPs, similar to www.gob.mx i guess.

Issue #600, opened 2015-08-11
Ingest not commited if error?


mccalluc:
It seems that some errors are preventing the commit from happening. I think the behavior we want is for all but the problematic records to go in to the index, but that does not seem to be happening reliably.

mccalluc:
Resolved, mostly by #618.

PR #618, opened 2015-08-12
More ingest stats 606

### LINKED ISSUE #606: More informative summary after ingest
>
How hard would it be to just add
- Total records processed
- Total updated (with %)
- Total skipped (with %)
- Total errors (with %)

mccalluc:
@afred: I think this does what you want? Fix #606.


mccalluc:
Also, fix #600: Old behavior was not to commit the entire collection if there was an error in any single document. This loosens that up: Any single document which validates should be committed.

PR #619, opened 2015-08-13
New map

mccalluc:
@caseyedavis12 : This is what it looks like; pins still need to be moved to Alaska. 
screen shot 2015-08-12 at 8 20 08 pm
Let me know if there is feedback.


mccalluc:
@afred: this is ready to go, if it looks good to you.

PR #629, opened 2015-08-14
Access facet tweaks 625

mccalluc:


afred:
@mccalluc please review change in catalog_spec.rb and make sure it's correct. Merge away.

mccalluc:
@afred: yeah: 4 -> 3 is correct: we're only going to have 3 options there. (Removing on-site-only is the other part of this commit.)

PR #634, opened 2015-08-17
Rubocop cleanup

mccalluc:
@afred: could you review and merge this before we start a lot of new branches? It had just been a while since I had cleaned up the code. This is also a newer version of rubocop, so that's part of it.


afred:
rebased.

afred:
@mccalluc do you see particular value in testing against the error output messages? It causes a tight coupling as highlighted in the 3rd commit of this PR.

mccalluc:
@afred : You're right. I think just checking the first line would be better. (2/3 are just one line, in any event.) Filed: #635.

PR #584, opened 2015-08-06
Browse the Collection on Homepage

Hammicakes:
-updated the _browse_list.html.erb. removed some formatting, outputs
row and list.
-added some custom jQuery in custom.js

PR #637, opened 2015-08-18
Bring back Hammicake's styling of the browse list.

mccalluc:
Revive portions of #584. Looks much better. 
screen shot 2015-08-18 at 1 21 49 pm


mccalluc:
Now looks like:
screen shot 2015-08-19 at 4 54 26 pm

Issue #554, opened 2015-07-30
State facets in side bar


mccalluc:
Follow up to #547: We want a limit-by-state facet.
- Easiest is just to remove the special casing that was in that PR
- but want to look at the possibility of combining with institutions list: having a header for the state, and then all the institutions under that.

mccalluc:
@caseyedavis12 / @sroosa : Would something like the mock-up below be worth any extra effort? (Alphabetical by state, etc, in the real thing.) I've worried for a while about how usable the long list of organizations is, and this addresses that. On the other hand, just dropping in a new facet would be trivial, so I'm reluctant to invest the time unless folks think it would be much better.
![screen shot 2015-08-04 at 10 59 29 am](https://cloud.githubusercontent.com/assets/730388/9063930/037a13e2-3a98-11e5-9ebe-6f6797bda480.png)

mccalluc:
done.

PR #646, opened 2015-08-19
State facet 554

mccalluc:
Put state facet as headings in the org facet. If we were to do it separately
- It would be less clear what "state" is all about.
- The orgs list is harder to read. (As it has been.)

I want the facet add/remove behavior to be better, but this isn't intolerable.


mccalluc:
(Towards #554.)

mccalluc:
Folks supported this idea. Link to state facet, but maybe exclude the state facet when counting stations?

mccalluc:
@afred: I think this is what we want: Clicking on the state brings in all the orgs under that state. (In the back of my mind, I'm wondering now if the links from the org pages should work the same way, but someone else can file a bug if they are so lead.)

lib/aapb.rb
```diff
@@ -1,3 +1,60 @@
...
+    'VT' => 'Vermont',
+    'VA' => 'Virginia',
+    'WA' => 'Washington',
+    'WV' => 'West Virginia',
+    'WI' => 'Wisconsin',
+    'WY' => 'Wyoming'
+  }
+  STATES.default = 'undefined' 
+  # Without a default we get a cryptic error message
+  # about comparing array to array for the sort.
```
afred: @mccalluc does this cryptic error come from ruby or javascript?

mccalluc: Ruby, but this will be moot, because this list is going to go away again with the NAEB stuff: I'm adding a state.rb and states.yml.

PR #648, opened 2015-08-20
We are not using search_field, and should error if we do see it.

mccalluc:


afred:
@mccalluc there was a conflict in `CatalogController#index`. I wasn't 100% on which bit to keep, so I merged instead of rebased.

Now it's failing [here](https://github.com/WGBH/AAPB2/blob/f5ca9acaf65ff4bd3a8eb3d7dfdba04bd9b0903b/spec/features/catalog_spec.rb#L191-L193). I'm digging into it, but I assume that you understand what this test is supposed to be doing? Do you have a quick fix?

mccalluc:
(I'm getting more errors than travis: Something is screwy with my checkout out. Hold on...)

afred:
@mccalluc the merge will have re-shuffled the history, so you may have to delete your local branch and re-checkout.

I think I may have found it though, [here](https://github.com/WGBH/AAPB2/blob/f5ca9acaf65ff4bd3a8eb3d7dfdba04bd9b0903b/app/controllers/catalog_controller.rb#L151-L156). Forcing this to be `PBCore::PUBLIC_ACCESS` passes the test.

afred:
^^ yes, on my local machine, it's saying i'm "onsite" to the access is going to `PBCore::PROTECTED_ACCESS`.

afred:
@mccalluc although, I'm pretty sure I saw the tests passing for this PR before doing the merge.

afred:
@mccalluc so it looks like the test is expecting the user not to be "on-site", which is supposed to add the `PBCore::PUBLIC_ACCESS` flag to the query string. Instead, when testing from localhost (and presumably Travis), the current use is classified a being "on-site", so the `PBCore::PROTECTED_ACCESS` flag is getting set instead.

What I'm unsure of, is whether this test is expecting the correct behavior, or if it needs changed.

mccalluc:
I'm trying to understand why the "Nova; Gratuitous Explosions" record doesn't count as public...

mccalluc:
@afred: One of the new things that came in with the merge was different default access filters depending on whether you're onsite or not... but I did it wrong, restricting on-site users to only PROTECTED items, when what I want is DIGITIZED. 

(DIGITIZED = PUBLIC + PROTECTED)

PR #661, opened 2015-08-26
Four more push-pins, and consistent query behavior

### LINKED ISSUE #651: city links on map are just anchors: need paths

### LINKED ISSUE #652: NAEB infrastructure
- new dots on map (specified by Casie)
- `states.yml` which lets us supply an extra blurb at the top the state section.
- If there is only a blurb, but with no actual stations, do not show the 'show all' link in the upper right.
- Casie will supply the text for the blurb.

mccalluc:
Fix #651 / Fix #652


mccalluc:
hmm... not getting the same error locally.

app/views/organizations/index.html.erb
```diff
@@ -15,25 +13,33 @@
   <br/>
   <br/>
 
-  <% state_list.each do |state,orgs| %>
-    <a name="<%= state %>"></a>
+  <% @states.select { |state| state.organizations || state.blurb_html }.each do |state| %>
```
mccalluc: why not all states?

mccalluc:
Audio content from [call letters] in [state] on participating orgs page

PR #671, opened 2015-09-01
Update tite, narrative and abstract

caseyedavis12:
Alan gave feedback on narrative and abstract. Also changing title. Do I need to change it elsewhere, such as in the breadcrumbs?


mccalluc:
@caseyedavis12 : no change needed elsewhere. I'll merge this.

PR #674, opened 2015-09-01
Turn on the ORR.

mccalluc:


mccalluc:
Hold this until WGBH records have been reindexed with more conservative privs. (Could merge now and wait on the redeploy, but that might make it messier if we had to pull it back.)

PR #684, opened 2015-09-04
Curated exhibits page

Hammicakes:
Updated for "View in Gallery" & "Featured" buttons.


mccalluc:
This is not in a state where it can be merged. Any code that we put in master needs to be ready to go on the live site, and that just isn't the case here. Also, this is far enough behind master that the merging would probably be painful.

The styling of the buttons looks good though, and I'll pull that out into a new PR.

PR #685, opened 2015-09-04
Grid Layout

Hammicakes:
Stripped .galley tags and Flexbox out of grid layout.

Updated javascript to assit in creating row with Bootstrap columns.

### DIFF
--- a/app/assets/javascripts/custom.jquery.js
+++ b/app/assets/javascripts/custom.jquery.js
@@ -1,0 +1,45 @@
+// custom.jquery.js
+
+$(document).ready(function () {
+
+	//check window size and inserts an <hr> appropriately
+	if ($(window).width() < 991) {
+		$("#documents.row .record:nth-of-type(even)").after("<hr>");
+	}
+	else {
+		$("hr").remove();
+		$("#documents.row .record:nth-of-type(3n)").after("<hr>");
+	}
+
+	//Same thing as about but accounts for window resize
+	var resizeId;
+	$(window).resize(function() {
+	  clearTimeout(resizeId);
+	  resizeId = setTimeout(doneResizing, 200);
+	});
+
+	function doneResizing() {
+		if ($(window).width() < 991) {
+			$("hr").remove();
+			$("#documents.row .record:nth-of-type(even)").after("<hr>");
+		}
+		else {
+			$("hr").remove();
+	    $("#documents.row .record:nth-of-type(3n)").after("<hr>");
+		}
+	}
+
+
+});
+
+// Listen for Button CLicks
+$(document).ajaxSuccess(function() {
+	//check window size and inserts an <hr> appropriately
+	if ($(window).width() < 991) {
+		$("#documents.row .record:nth-of-type(even)").after("<hr>");
+	}
+	else {
+		$("hr").remove();
+		$("#documents.row .record:nth-of-type(3n)").after("<hr>");
+	}
+});
--- a/app/assets/javascripts/custom.jquery.js
+++ b/app/assets/javascripts/custom.jquery.js
@@ -2,44 +2,26 @@
 
 $(document).ready(function () {
 
-	//check window size and inserts an <hr> appropriately
-	if ($(window).width() < 991) {
-		$("#documents.row .record:nth-of-type(even)").after("<hr>");
-	}
-	else {
-		$("hr").remove();
-		$("#documents.row .record:nth-of-type(3n)").after("<hr>");
-	}
 
-	//Same thing as about but accounts for window resize
+  function doneResizing() {
+    if ($(window).width() < 991) {
+      $("#documents.row hr").remove();
+      $("#documents.row .record:nth-of-type(even)").after("<hr>");
+    }
+    else {
+      $("#documents.row hr").remove();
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/custom.jquery.js
```diff
@@ -0,0 +1,45 @@
...
+$(document).ready(function () {
+
+	//check window size and inserts an <hr> appropriately
+	if ($(window).width() < 991) {
+		$("#documents.row .record:nth-of-type(even)").after("<hr>");
+	}
+	else {
+		$("hr").remove();
+		$("#documents.row .record:nth-of-type(3n)").after("<hr>");
+	}
```
mccalluc: You've defined all of this in the function below, so you should just be able to use the function. Copy and paste will be less maintainable over time.

app/assets/javascripts/custom.jquery.js
```diff
@@ -0,0 +1,45 @@
...
+	  resizeId = setTimeout(doneResizing, 200);
+	});
+
+	function doneResizing() {
+		if ($(window).width() < 991) {
+			$("hr").remove();
+			$("#documents.row .record:nth-of-type(even)").after("<hr>");
+		}
+		else {
+			$("hr").remove();
```
mccalluc: Could you tighten up this pattern? Maybe someone decides to use an hr somewhere else on the page: it will be frustrating when they keep on disappearing.

app/assets/javascripts/custom.jquery.js
```diff
@@ -0,0 +1,45 @@
...
+$(document).ajaxSuccess(function() {
+	//check window size and inserts an <hr> appropriately
+	if ($(window).width() < 991) {
+		$("#documents.row .record:nth-of-type(even)").after("<hr>");
+	}
+	else {
+		$("hr").remove();
+		$("#documents.row .record:nth-of-type(3n)").after("<hr>");
+	}
+});
```
mccalluc: again, use doneResizing rather than copy and paste. Could you describe what events this is necessary for? The site doesn't seem that ajax-y to me.

app/assets/stylesheets/application.css
```diff
@@ -375,14 +375,45 @@ hr {
 /* -------------------- */
 
 /* This is a record display in a search result*/
+#documents.row {
+	/*border-color:#DDD;
+	border-style:solid;
+	border-width:1px 0 0 1px;*/
+}
```
mccalluc: Here and below there's a lot of CSS which is commented out... do we need it hanging around for some reason?

mccalluc:
@Hammicakes : also note the merge conflicts: This is 120 commits behind master, so it's not that surprising that something conflicts. With all the possible CSS interactions, I feel less than comfortable resolving these conflicts on my end. Options:
- Merge the changes from master into this branch.
- Rebase this branch on top of master.
- or, what may be easiest, just abandon this branch, make a new one, and make your edits there.

afred:
@mccalluc @Hammicakes i rebased this and tests are passing. But the JS for inserting/removing `<hr>` on window resize is broke, due to ineffective CSS selectors. Can either of you describe, in english, what the idea is here? From the code it looks like...
  1. when window is less than 991px wide, insert an `<hr>` after every other record
  1. when window is more than 991px wide, insert an `<hr>` after every third record

This logic doesn't make much sense to me. Am i missing something?

afred:
^ derp... i just realized you need to be in grid layout

afred:
@mccalluc ok... so i played around with it in grid view, and it's inserting/removing the `<hr>` accordingly. But i'm not sure if the `<hr>` is just used for positioning other floated elements, or if it's supposed to be seen. Currently, it's not visible, and it appears that is by design. But I'm not sure what we were going for here. In any event, the page looks fine to me.

mccalluc:
@afred:
>
    when window is less than 991px wide, insert an <hr> after every other record
    when window is more than 991px wide, insert an <hr> after every third record

So I think this is really about fixing the behavior on safari, where they weren't wrapping within the screen. Narrower widths, 2 columns, Wider, 3 columns. Not sure if that helps.

afred:
@mccalluc ah.. well in that case, adding the `<hr>` is not working in Safari -- at least not to create rows of 2 instead of three.

In the image below, you can see a narrow Safari window with a row of 3 items, and to the right, you can see the DOM inspector, where there is an `<hr>` element after the first two `<div class="record">` elements.

image

So the JS is functionally working, but the `<hr>` is not having the desired effect.

afred:
I consider this to be a pretty minor bug. There appears to be a lot of style changes, so I'm assuming this PR does something more than just the Safari row-size fix. I'm in favor of merging this, and isolating the Safari resize in another issue.

mccalluc:
I remain uncomfortable with this. We're not sure what it's doing, and the one thing that it claims to do
> Updated javascript to assit in creating row with Bootstrap columns

doesn't work.

afred:
@mccalluc ah... i didn't read your prev comment closely enough re: Safari not wrapping at all. This is definitely the case. Here's a screenshot of Safari with fixture data ingested:
image

mccalluc:
Going to close this PR since it doesn't do what we need it to do. The branch can stay around for reference.

Issue #681, opened 2015-09-03
no-results suggestion not working on production


mccalluc:
`http://localhost:3000/catalog?f[access_types][]=online&f[asset_type][]=Raw+Footage&q=` ->
```html
<a href="/catalog?f[access_types][]=all&amp;f[asset_type][]=Raw+Footage&amp;q=">
  searching all records
</a>
```

but ...

`http://americanarchive.org/catalog?f[access_types][]=online&f[asset_type][]=Raw+Footage&q=` ->
```html
<a href="/catalog?f%5Baccess_types%5D%5B%5D=online&amp;f%5Basset_type%5D%5B%5D=Raw+Footage&amp;q=">
  searching all records
</a>
```

Regex based replacement in _zero_results.erb isn't handling this correctly.

(It might be that the first URLs are more different than they appear? In any event, this needs to be fixed.)

mccalluc:
fixed.

PR #687, opened 2015-09-04
More robust no-results logic

mccalluc:
I tried to reproduce this on localhost, and I tried producing a failing test, but I could not. Annoyed, but I think this is better, regardless.


mccalluc:
fix  #681

PR #693, opened 2015-09-10
Station history exhibit

sroosa:
I'm definitely going to need to add more to this exhibit, but this part can be merged now.


mccalluc:
@sroosa: Rather than trying to fix the conflicts here, I created #695.

PR #696, opened 2015-09-11
Intl no access 642

### LINKED ISSUE #642: International no-access message
Splitting from #577: 
- What message do we give to international users when we don't show them the video?
- Should this be in place of / in addition to / or optional if there is already a message because it is onside-only or private.

mccalluc:
Fix #642
- @afred: Please review.
- @caseyedavis12 : um, I used a different message: This just seems really unclear to me.
 
> Content in the ORR is not currently available in your geography.

  - Folks just landing on this page are not going to know what "ORR" means.
  - "currently": I would wonder if this is a temporary outage or something longer term.
  - I would have no idea what my "geography" is.

### DIFF
--- a/app/models/pb_core.rb
+++ b/app/models/pb_core.rb
@@ -241,7 +241,7 @@ class PBCore # rubocop:disable Metrics/ClassLength
                :rights_code, :access_level, :access_types,
                :organization_pbcore_name, # internal string; not in UI
                :titles_sort, :title, # Covered by #titles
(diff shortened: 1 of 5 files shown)


caseyedavis12:
@mccalluc: your message is fine, but please reinsert "currently" or "at this time." This was requested by the legal counsel since eventually we may be able to open up the content to other countries.

afred:
not crazy about the `alias_method_chain` pattern.. but i tried an alternative and it didn't look much better. Merging.

PR #700, opened 2015-09-11
If media type is "other", just use the audio player

### LINKED ISSUE #697: WYSO digitized assets not available on public website
If you go to americanarchive.org and look at WYSO content, many of the records have the word "true" where the audio player should be. Example: http://americanarchive.org/catalog/cpb-aacip_27-gx44q7r549 

Here is the asset in the AMS which has a Sony Ci id: https://ams.americanarchive.org/records/details/4893209

mccalluc:
@caseyedavis12 : Is this ok, or should ingest fail if we have ci id, and "other" media type, as we do here? http://americanarchive.org/catalog/cpb-aacip_27-gx44q7r549.pbcore:
```xml
<pbcoreIdentifier source="Sony Ci">ad23e7dbefd346c49fa37de4c26f1515</pbcoreIdentifier>
...
<instantiationMediaType>other</instantiationMediaType>
```

I'm comfortable with this: downside:
- Facet info wrong; icon missing

upside:
- save a lot of recataloging work

Fix #697. @afred : please review.

### DIFF
--- a/app/views/catalog/_player.erb
+++ b/app/views/catalog/_player.erb
@@ -8,31 +8,27 @@
 </script>
 
 <%= 
-  if @pbcore.video? || @pbcore.audio?
-    @pbcore.media_srcs.map do |media_src|
-      media = @pbcore.video? ? 'video' : 'audio'
-      content_tag(media, 'class' => 'video-js vjs-default-skin', 
-                  controls: true,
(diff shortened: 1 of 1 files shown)


caseyedavis12:
@mccalluc this is OK

PR #706, opened 2015-09-14
Caption 699

### LINKED ISSUE #699: Captions just not showing up
I felt like when I had stubbed out captions before I had at least gotten something on the screen, but now I'm not even able to do that. Have tried both 
- SRT from LoC and 
- VTT that works for other demos on the web.

 In the browser console, I can see that the request for the caption file is working, but it never shows up in the player. If it were a cross site security problem, it shouldn't be loading at all.

Strategy: start with a plain VideoJS player just on local html and see if that works.

mccalluc:
Fix #699.
@afred: Please review. This got more complicated than I was thinking at first, but I think it's the right direction. The center of the complexity is `to_solr`: This is only called at index time, so I think it's the right place to ping s3 so that we can tell (1) whether the captions even exist, and (2) what the content is, but only for the purpose of indexing.

I think this will be of increasing utility as we have transcript-y things which don't belong in the pbcore, but which should be indexed for retrieval.

### DIFF
--- a/app/models/pb_core.rb
+++ b/app/models/pb_core.rb
@@ -96,13 +96,11 @@ class PBCore # rubocop:disable Metrics/ClassLength
   def media_srcs
     @media_srcs ||= (1..ci_ids.count).map { |part| "/media/#{id}?part=#{part}" }
   end
+  CAPTIONS_ANNOTATION = 'Captions URL'
   def captions_src
-    @captions_src ||= begin
-      s3_id = id.gsub('_','-')
-      base = 'https://s3.amazonaws.com/americanarchive.org/captions'
-      "#{base}/#{s3_id}/#{s3_id}.srt1.srt" if video?
-    end
-    # TODO: "CC" icon will show for all videos, even if caption isn't actually available.
+    @captions_src ||= xpath("/*/pbcoreAnnotation[@annotationType='#{CAPTIONS_ANNOTATION}']")
+  rescue NoMatchError
+    nil
   end
   def img_src
     @img_src ||=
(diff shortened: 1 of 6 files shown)


app/models/pb_core.rb
```diff
@@ -199,12 +201,40 @@ def access_types
...
+          caption_url + '</pbcoreAnnotation>'
+        )
+    end
+    
+    xml_holder = []
+    doc_with_caption_flag.write(xml_holder)
     {
       'id' => id,
-      'xml' => @xml,
+      'xml' => xml_holder.join(),
```
afred: @mccalluc can you explain the purpose of `xml_holder`?

mccalluc: @afred: That and the lines above are how you render REXML without creating an explicit renderer... but the next commit is better: Singleton `Formatter` since there are actually two places we render xml, and it'd be good if it were the same in both.

app/models/pb_core.rb
```diff
@@ -199,12 +202,38 @@ def access_types
...
+      pre_existing = REXML::XPath.match(doc_with_caption_flag, "//pbcoreAnnotation[@annotationType='#{CAPTIONS_ANNOTATION}']").first
+      if pre_existing
+        pre_existing.parent.elements.delete(pre_existing)
+      end
+      caption_body = caption_response.body
+      REXML::XPath.match(doc_with_caption_flag, '/*/pbcoreInstantiation').last.next_sibling.next_sibling = 
+        REXML::Document.new(
+          "<pbcoreAnnotation annotationType='#{CAPTIONS_ANNOTATION}'>" +
+          caption_url + '</pbcoreAnnotation>'
+        )
```
mccalluc: This would be a bit better constructed programmatically, rather than from a string...

spec/models/pb_core_spec.rb
```diff
@@ -90,8 +90,6 @@
     end
 
     describe 'full' do
-      pbc = PBCore.new(pbc_xml)
-
```
mccalluc: Just moved this declaration down closer to where it is used.

afred:
@mccalluc :+1: for refactoring. Looks good.

PR #716, opened 2015-09-15
No map responsive 710

### LINKED ISSUE #710: Don't do maps on mobile/narrow screens
For some reason we don't understand, map is not rendering on iphone. (Can't reproduce on desktop.)

Rather than worrying over it, let's just take maps out of the narrow responsive view: the states are too small to click on, in any case.

mccalluc:
Fix #710. @afred: please review and sanity check... We're sweeping it under the rug a little bit, but with limited real estate there are good reasons to skip the map.


afred:
@mccalluc good call

PR #724, opened 2015-09-16
Move magazine, news report, and performance, as requested. 

### LINKED ISSUE #722: Some terms appearing as topic rather than genre
@mccalluc I just noticed this. Magazine, News Report and Performance should be on the genre list, not the topics list. I think this is probably something that gets sorted on ingest, so don't worry about it right now, as the term is still appearing somewhere. Are we planning on doing a full reingest at some point in the future? Can we make the fix then?

mccalluc:
No effect till next reingest.
Fix #722.

### DIFF
--- a/config/vocab-maps/genre-type-map.yml
+++ b/config/vocab-maps/genre-type-map.yml
@@ -21,7 +21,7 @@
   Interview: Interview
   Magazine: Magazine
   Mini: Miniseries
-  News Report: News Report
+  News: News Report
   #Performance in a Studio: Performance in a Studio
   #Performance for a Live Audience: Performance for a Live Audience
   Performance: Performance
--- a/config/vocab-maps/genre-type-map.yml
+++ b/config/vocab-maps/genre-type-map.yml
@@ -21,9 +21,7 @@
   Interview: Interview
   Magazine: Magazine
   Mini: Miniseries
-  News: News Report
-  #Performance in a Studio: Performance in a Studio
-  #Performance for a Live Audience: Performance for a Live Audience
+  News Report: News Report
   Performance: Performance
   Promo: Promo
   Public Service Announcement: Public Service Announcement
--- a/config/vocab-maps/genre-type-map.yml
+++ b/config/vocab-maps/genre-type-map.yml
@@ -19,7 +19,12 @@
   How to: Instructional
   How-to: Instructional
   Interview: Interview
+  Magazine: Magazine
   Mini: Miniseries
+  News Report: News Report
+  #Performance in a Studio: Performance in a Studio
+  #Performance for a Live Audience: Performance for a Live Audience
+  Performance: Performance
   Promo: Promo
   Public Service Announcement: Public Service Announcement
   PSA: Public Service Announcement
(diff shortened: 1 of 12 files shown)


mccalluc:
@sroosa : What should be the behavior for "Newsmagazine"... or more generally, if a term could map to either a topic or a genre? 
- Old behavior was to silently delete the element. 
- New behavior is to fail ingest. 
- Right behavior might be to default to one over the other?

sroosa:
We can have it default to Magazine. Thanks!

afred:
@mccalluc I'm trying to fix this test... stand by.

sroosa:
Both for newsmagazine would be great. But otherwise default to Format: Magazine.

mccalluc:
@afred please review. Having one term map to both a genre and topic would get messy: Right now that gets reported as an error. So
- "Newsmagazine" -> genre: "Magazine"
- "News" -> genre: "News Report"

sroosa:
@mccalluc: Can "News" still map to a Topic "News"? There are some other genres of content, like Documentaries, Magazines, etc. that still have a "news" topic. Thanks!

afred:
so here's my understanding of what's going on...

  - `<pbcoreGenre>Newsmagazine</pbcoreGenre>` now gets mapped to `<pbcoreGenre annotation="genre">Magazine</pbcoreGenre>`.
  - `<pbcoreGenre>News</pbcoreGenre>` was getting mapped to `<pbcoreGenre annotation="topic">News</pbcoreGenre>`, but is now getting mapped to `<pbcoreGenre annotation="genre">News Report</pbcoreGenre>`

is this correct @mccalluc?

mccalluc:
> Can "News" still map to a Topic "News"
@sroosa: The big problem is when one term that's coming in hits a match on both lists. Right now the code just puts on a label on it ... Having it generate a new term in the xml is a complexity I'd like to avoid.

But not impossible.

Another approach would be to consider this as a cataloging problem: "Newsmagazine" is just bad data because it matches on both lists, and it's just something that needs to be cleaned up.

So that's the problem space, as I understand it.

sroosa:
@mccalluc That is definitely where the problem is. In the records we've cataloged, those should be split out in to two terms, so each only matched on list.

mccalluc:
I'll try to get the machinery to generate two terms from the one. Hold on...

sroosa:
@mccalluc I think i misunderstood. I don't mind that it doesn't generate two terms. I just wanted to ensure that News would still be on the Topics list in general, and News Report would still be on the Genre list in general.

config/vocab-maps/genre-type-map.yml
```diff
@@ -19,7 +19,12 @@
   How to: Instructional
   How-to: Instructional
   Interview: Interview
+  Magazine: Magazine
   Mini: Miniseries
+  News: News Report
```
mccalluc: *just* "News Report" -> "News Report"

config/vocab-maps/topic-type-map.yml
```diff
@@ -37,14 +37,8 @@
...
   Local Communities: Local Communities
   Recorded Music: Recorded Music
-  #Performance in a Studio: Performance in a Studio
-  #Performance for a Live Audience: Performance for a Live Audience
-  Performance: Performance
   Music: Music
   Nature: Nature
-  Magazine: Magazine
-  News Report: News Report
-  News: News
```
mccalluc: *keep this*

mccalluc:
@afred: This is what I should have done in the first place. Please review.

PR #727, opened 2015-09-16
Md defaults

mccalluc:
@afred: please review. Cmless now has inheritance.


afred:
@mccalluc would you want to take the inherited stuff out of https://github.com/WGBH/AAPB2/blob/master/spec/fixtures/exhibits/parent/child/grandchild.md in order to test the inheritance?

PR #730, opened 2015-09-23
Editing sadies exhibit

sroosa:


mccalluc:
@sroosa : The arkansas item you removed was actually an item the tests were relying on. I've put it back, just so the tests will pass. We can do another PR to take it out again, but it's a little bit of work, and I'd want to be sure it's worth it.

PR #734, opened 2015-09-24
solr settings for captions were exactly backwards.

mccalluc:


mccalluc:
@afred, now that you're back, please review.

PR #742, opened 2015-10-07
Adds demo site as an allowed referer

afred:
Adds tests for User#aapb_referer?


afred:
@mccalluc please review

Issue #748, opened 2015-10-09
Errors from latest ingest: try to reproduce locally


mccalluc:
These produced rsolr error when trying to do an ingest on EC2: Try to reproduce locally.

[...]

PR #752, opened 2015-10-09
Ingest errors 748

mccalluc:
@afred: This fixes all the problems in #748. Let's hold off on closing until they've been reingested on the live server.


mccalluc:
(Thinking about it, the regex is a little bit of overkill: we don't actually need to worry about preserving new lines, since this whole thing is just being provided to solr for indexing. But it doesn't do any harm either.)

PR #757, opened 2015-10-13
Completely ad-hoc, but looks a lot better.

### LINKED ISSUE #384: Grid view in Safari
When a user conducts a search in Safari, the grid view appears horizontally across the page, requiring the user to scroll left and right across the page. The thumbnails do not appear stacked in columns of three as in other browsers. 

Done when:
Search results (grid view) in Safari behaves the same way as Chrome or Firefox

mccalluc:
@afred: please review. Fixes #384. CSS can be moved out of the ERB, but for a review, thought it made things more clear to have it in line.

(Does the `.gallery` class do something useful and we'll regret removing it someday? I really don't know.)

### DIFF
--- a/app/assets/stylesheets/blacklight_overrides/_catalog.scss
+++ b/app/assets/stylesheets/blacklight_overrides/_catalog.scss
@@ -1,0 +1,14 @@
+
+$grid-item-height: 250px;
+
+#documents {
+  .document {
+    border-bottom: none;
+    height: $grid-item-height;
+  }
+
+  .thumbnail {
+    height: $grid-item-height;
+    overflow: hidden;
+  }
+}
--- a/app/assets/stylesheets/blacklight_overrides.scss
+++ b/app/assets/stylesheets/blacklight_overrides.scss
@@ -1,0 +1,3 @@
+// Includsion file for all of blacklight overrides.
+
+@import 'blacklight_overrides/_catalog.scss';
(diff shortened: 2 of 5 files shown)


afred:
@mccalluc using `height: 15em;` is showing like this in Safari

image

Not the end of the world. I will poke around a little, but if it's a pain, I can live with what we have for now.

afred:
hmm... same behavior actually is happening in Chrome...

image

afred:
@mccalluc ready for review

Per your question above, `.gallery` class is what adds all of the flex-grid stuff. It seems really nice actually, and time-saving, so if we can get Safari to work better with it, then I would be in favor of returning to it.

Right now, the only difference after removing `.gallery` that I've noticed is that it will default to 4 columns (instead of 3) for the large breakpoint

app/assets/stylesheets/blacklight_overrides/_catalog.scss
```diff
@@ -0,0 +1,14 @@
+
+$grid-item-height: 250px;
+
+#documents {
```
mccalluc: @afred: do we want to add anything else here to make it clear that this is only about the gallery view? We're sure `.document` and `.thumbnail` aren't used anywhere else?

afred:
@mccalluc good idea re: earlier comment about being more specific in applying styles to `.document` and `.thumbnail` classes. I added a specific class to act as a replacement for the now unused `.gallery` class. please review

PR #767, opened 2015-10-20
Civil rights 764

mccalluc:
@caseyedavis12 : Merge this and then continue. 

(FWIW: Git abhors a vacuum: You can't actually create an empty directory.)


mccalluc:
looking into errors...

PR #768, opened 2015-10-20
Exhibits 765

### LINKED ISSUE #765: Enable "Curated Exhibits"
**DO NOT DEPLOY UNTIL Monday, 10/26**

**Done when:**
  * change any verbiage for "exhibition" to "exhibit"
  * section is added with title "Curated Exhibitions"
    * placement: below the map
  * contains rotating gallery -- may not need to rotate, since there are only 3 exhibitions
    * 3 items in rotating gallery contain the signature image (first img in markdown)
  * Add "Exhibitions" to main nav and floating nav for home page
  * Add "Curated Exhibitions" to the "about" menu, linking to /exhibits

mccalluc:
**DO NOT DEPLOY UNTIL Monday, 10/26**

fix #765, except for the last part where I'm confused exactly where the link goes.


caseyedavis12:
@mccalluc: We do not want sub-exhibits to appear in the carousel.



________________________________

mccalluc:
@caseyedavis12 : yep, later commit fixed that.

mccalluc:
@afred: We won't have another interim release, so this could be merged, I think.

app/views/home/_exhibits_carousel.erb
```diff
@@ -1,18 +1,15 @@
...
                     view_all_href: '/exhibits'
                    } do |item| %>
   <a href="<%= "/exhibits/#{item.path}" %>" class="list-group-item" >
-    <img src="<%= item.thumbnail_url %>" title="<%= item.name %>" alt="<%= item.name %>" />
+    <img src="<%= item.thumbnail_url %>" title="<%= item.title %>" alt="<%= item.title %>" />
     <span class="carousel-caption">
-      <h3 class="list-group-item-heading"><%= item.name %></h3>
+      <h3 class="list-group-item-heading"><%= item.title %></h3>
     </span>
   </a>
```
afred: @mccalluc is there a reason this markup is being passed as a block to `render`, instead of being handled by a partial, say something like, `home/_carousel_item.html.erb`?

mccalluc: Well, not a change with this commit, but I guess I didn't go that way because making a partial for the content of a loop isn't my default: 6 lines isn't overwhelming in this context, and it's not being used elsewhere.

afred: well, it's an interesting approach -- sort of dependency injection for view logic. I think we stay a little more flexible and modular if we stick with using partials. But this can merge.

PR #788, opened 2015-10-26
774 exhibit style tweaks

afred:
closes #744 

@mccalluc please review

**UPDATE:** @mccalluc This PR had other stuff in it that we actually _do_ want right now. So please review, and refer to images below.

### DIFF
--- a/app/assets/stylesheets/exhibits.scss
+++ b/app/assets/stylesheets/exhibits.scss
@@ -1,0 +1,12 @@
+.exhibit-index {
+
+  $grid-item-height: 350px;
+
+  .thumbnail {
+    height: $grid-item-height;
+
+    @media (max-width: 992px) {
+      text-align: center;
+    }
+  }
+}
--- a/app/assets/stylesheets/exhibits.scss
+++ b/app/assets/stylesheets/exhibits.scss
@@ -1,9 +1,7 @@
 .exhibit-index {
-
-  $grid-item-height: 350px;
-
+  
   .thumbnail {
-    height: $grid-item-height;
+    height: 350px;
 
     @media (max-width: 992px) {
       text-align: center;
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -13,14 +13,14 @@ class CatalogController < ApplicationController
     ## See also SolrHelper#solr_search_params
     config.default_solr_params = {
       qt: 'search',
-      rows: 10
+      rows: 12
     }
 
     # solr path which will be added to solr base url before the other solr params.
     # config.solr_path = 'select'
 
     # items to show per page, each number in the array represent another option to choose from.
-    # config.per_page = [10,20,50,100]
+    config.per_page = [12,30,60,90]
 
     ## Default parameters to send on single-document requests to Solr.
     ## These settings are the Blackligt defaults (see SolrHelper#solr_doc_params) or
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -150,6 +150,11 @@ class CatalogController < ApplicationController
   end
 
   def index
+
+    # If we are looking at search results for a particular exhibit, then fetch
+    # the exhibit for additional display logic.
+    @exhibit = exhibit_from_url
+
     if !params[:f] || !params[:f][:access_types]
       base_query = params.except(:action, :controller).to_query
       access = if current_user.onsite?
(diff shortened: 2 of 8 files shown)


afred:
@mccalluc **DON'T MERGE!!** the main requirement that lead to this refactor is no longer needed.

However, also **PLEASE DON'T CLOSE** because in general, I think some of this refactor is still valuable.

Going to remove from the "Launch" milestone for now, and we can come back to it later.

afred:
**Regular search results:**
image

caseyedavis12:
@afred looks fine to me. Thanks!

afred:
**Show All Exhibit Items**
image

caseyedavis12:
@afred looks good to me

afred:
**Show all Exhibits**
image

app/views/catalog/_search_header.erb
```diff
@@ -1,14 +1,6 @@
...
-      exhibit = Exhibit.find_by_path(path)
-      %>
-        <%= link_to(exhibit.title, "/exhibits/#{exhibit.path}") %>:
-        <%= sanitize(Exhibit.find_by_path(path).summary_html, tags: []) %>
-      <%
-    end if exhibit_paths
-  end
-%>
+<% if @exhibit %>
+  <%= render partial: 'exhibits/show_all_items_header', locals: {exhibit: @exhibit } %>
```
mccalluc: In general, do we prefer passing variables to partials rather than relying on fields from the instance?

app/controllers/catalog_controller.rb
```diff
@@ -182,4 +187,18 @@ def show
...
+
+  # Returns exactly 1 Exhibit object given a path to the exhibit in the params
+  def exhibit_from_url
+    # Despite 'exhibit' field being multi-valued in solrconfig.xml, we're only
+    # returning the first exhibit from the URL we currently only allow users to
+    # select 1 exhibit in the UI, via the 'Show all items' link on the exhibit
+    # pages are Cmless pages.
+    if params['f'] && params['f']['exhibits'] && !params['f']['exhibits'].empty?
+      path = params['f']['exhibits'].first
+      return Exhibit.find_by_path(path) rescue nil
```
mccalluc: Are we ever really interested in seeing errors from this? All the checking seems excessive, if we just have a `rescue` anyway. Could we just simplify it to:
```ruby
def exhibit_from_url
  Exhibit.find_by_path(params['f']['exhibits'])
rescue
  nil
end
```

app/views/exhibits/_show_all_items_header.html.erb
```diff
@@ -0,0 +1,2 @@
+<%= link_to(exhibit.title, "/exhibits/#{exhibit.path}") %>
+<%= sanitize(exhibit.summary_html, tags: []) %>
```
mccalluc: hmm: Trying to remember why the sanitize is necessary?

app/controllers/catalog_controller.rb
```diff
@@ -13,14 +13,14 @@ class CatalogController < ApplicationController
...
-      rows: 10
+      rows: 12
     }
 
     # solr path which will be added to solr base url before the other solr params.
     # config.solr_path = 'select'
 
     # items to show per page, each number in the array represent another option to choose from.
-    # config.per_page = [10,20,50,100]
+    config.per_page = [12,30,60,90]
```
mccalluc: Here and above, does it just need to be a multiple of 3, or is it also important to be a multiple of 2? Comment explaining might be good.

app/views/catalog/_search_header.erb
afred: variables to partials is better than field from the instance, yes. It more clearly defines the interface to the partial.

app/assets/stylesheets/exhibits.scss
```diff
@@ -0,0 +1,10 @@
+.exhibit-index {
+  
+  .thumbnail {
+    height: 350px;
+
+    @media (max-width: 992px) {
```
mccalluc: Would you want to use one of the bootstrap variables here? And having the text centered when smaller is actually what we want, in any case?

app/controllers/catalog_controller.rb
afred: the above would conflate errors in accessing `params['f']['exhibits']` with any error thrown by `Exhibit.find_by_path`, which I think is too assuming. We want to ignore any `undefined method [] for nil class` errors, but we want to catch anything thrown by `Exhbiit.find_by_path`.

app/controllers/catalog_controller.rb
afred: multiples of 3 is what I was going for, i.e. square it off for the most common view port, which I assume is medium or larger. Round number look better in the "X per page" drop down for page size.

app/assets/stylesheets/exhibits.scss
afred: bootstrap variables aren't being used elsewhere, at least not that I could see. I figured we could refactor that later. If I'm wrong, then yes, we eventually want to be using _some_ kind of sass variable here.

and yes, the left-aligned text looks really goofy when you get smaller -- but only because this is still using flex grid, which centers the thumbnails per gallery item. If we disable the flex grid for this view, then let's make that a separate ticket.

afred:
@mccalluc this is ready for merge pending a :+1: from you.

PR #800, opened 2015-10-30
Records from Mississippi should be given clean IDs now.

### LINKED ISSUE #798: Mississippi Public Broadcasting thumbnails not appearing
http://americanarchive.org/catalog?f%5Borganization%5D%5B%5D=Mississippi+Public+Broadcasting+%28MS%29&f[access_types][]=digitized

mccalluc:
Fix #798. @afred?

Thought about moving the `.tr('/_', '_/')` out of PBCore. If we were starting from scratch I would put everything in Cleaner, but at this point it would mean changing each of the clean fixtures, and it would make all of the live data is slightly out of date... 

If we do it this way, instead, I think the right way to think about it is that there is the real AACIP ID which has a slash, or ought to, and the cleaner fixes it up that way, and there is a distinct AAPB ID for website use which has an underscore instead. Seem ok?


afred:
This is ok for now, but i support a refactor to move manipulation of ID string into `Cleaner`, fix the fixtures, and re-ingest data. Good house keeping, and less confusion for future us.

PR #801, opened 2015-10-30
End result looks good to me. Fix #797?

### LINKED ISSUE #797: Grid view for audio - no metadata appearing
We didn't notice that the grid view for audio cuts off all text, since the image is 250 pixels. We may need to increase the size of the grid view again.

mccalluc:
@afred? Maybe you want more of the actual math inline? Assuming the end result is what we want?

screen shot 2015-10-30 at 4 21 00 pm


caseyedavis12:
@mccalluc: Awesome!


cc: @afred

Issue #735, opened 2015-09-25
Homepage when narrow bounces around


mccalluc:
When the homepage is narrow, lower content on the page bounces around because carousel content changes height: Audio logos are taller than the thumbnails.

![screen shot 2015-09-25 at 12 07 33 pm](https://cloud.githubusercontent.com/assets/730388/10105702/27ad4018-637e-11e5-9268-3f7406be6ebb.png)

mccalluc:
@afred: After mucking with CSS for a while, I feel like the best option may just to make crop them:
```
convert public/thumbs/audio-digitized.jpg -crop 300x225+0+52 /tmp/crop.jpg
```
(`+52` so the icon and text are vertically centered.) The next question in my head is whether we really need/want the 1x1 images anywhere: I feel like we keep running into these cases (like the gallery) where having to deal with two different sizes does not make our lives easier, and even in list view results, I don't think the square images will me missed. Thoughts?

afred:
@mccalluc I agree that the generic thumbnails should probably be 4:3 to make them more similar to video thumbnails. I think this would make things appear more coherent, and in places like the home page, make responsiveness more predictable. IOW, if we don't have any square video thumbnails, then why have square generic thumbnails?

I'm also in favor of just getting the image sizes correct off the bat, rather than trying to resize them on the fly. The reasoning here is because images can be displayed with `<img>` or as backgrounds to block elements, and resizing is done differently in both cases, and getting them to resize in the same way without side effects can be tricky, as I found out with the search magnifying glass thing. I'd rather have thumbnail size be a design decision handed to developers as-is.

All that being said.. fwiw.. the home-page carousels respond pretty well on my phone, even with the 1:1 audio thumbnails. If by "jumping around" you mean that the content below it moves up and down ~20px or so.. i'm not sure that is a big problem. Maybe you're seeing something I'm not.

caseyedavis12:
@mccalluc and @afred: I don't have a problem with the "jumping around"

mccalluc:
Really, I'm not imagining this. Imagine you were trying to actually read the text in the next section.

![screen shot 2015-11-05 at 9 01 57 am](https://cloud.githubusercontent.com/assets/730388/10970523/328e7e2c-839d-11e5-9db6-7f5807e8f32f.png)

- More pronounced in tablet than phone, actually.
- Worst if you go from three squares to three rectangles

afred:
[...]

PR #805, opened 2015-11-02
Homepage CSS -> SCSS

mccalluc:
@afred: Wanted to start on #735... but I'd feel better if I was working in SCSS to minimize the risk of side effects, so tried to convert it without touching anything else first.
1. Is this a good idea, and necessary?
2. Do you see problems with the rendering with this, either on the home page, or on other, deeper pages?


afred:
:+1: for sass. Fwiw, SASS is a proper superset of CSS, therefore all valid CSS is also valid SASS.

mccalluc:
@afred: Did you actually check out how this renders? I'd actually be more surprised than not if this worked perfectly. It wasn't just changing the file name: I tried to get the internal organization right, too.

Issue #808, opened 2015-11-02
Capybara click-thru test of ORR agreement


mccalluc:
Back at 276b6e856 when the ORR was introduced there was also a test of the TOU agreement process, but at 465aa4f7b the part about clicking AGREE was removed: The Capybara user was already approved and was not seeing the TOU page. (For details: `git log -p -S AGREE spec`)

We need to get a test of the TOU back in place.

mccalluc:
Tested now with #810.

Issue #799, opened 2015-10-30
spec embed functionality


mccalluc:
@caseyedavis12 : Let us know when this needs to be delivered, and let's fill in the gaps.
- What does it look like on the embedder's end? Maybe:
![screen shot 2015-10-30 at 11 52 31 am](https://cloud.githubusercontent.com/assets/730388/10850827/02ccf158-7efe-11e5-89a6-8765a962a753.png)
- What does it look like on the embeddee's end? I would lean towards just the video, and a link below to the AAPB. Even including the title or other metadata could be a distraction: data-quality issues + having to deal with varying heights.
- Embeddee needs to click through same TOU, right?
- Don't think we can start video playback from an arbitrary point with Ci, but could be mistaken. That's ok? (Offering playback from a particular point also raises more UI questions.)
- No difference between video and audio embed.
- Only ORR will be embeddable: ie, if you're onsite and viewing something, we need to make sure that it's ORR before showing you the embed code.

caseyedavis12:
@mccalluc: If possible, could this be delivered by November 15?

  *   What does it look like on the embedder's end? Your screenshot looks good!

  *   What does it look like on the embeddee's end? I would lean towards just the video, and a link below to the AAPB. Even including the title or other metadata could be a distraction: data-quality issues + having to deal with varying heights. Video and link to the record on AAPB.
  *   Embeddee needs to click through same TOU, right? Yes
  *   Don't think we can start video playback from an arbitrary point with Ci, but could be mistaken. That's ok? (Offering playback from a particular point also raises more UI questions.) OK for now. We can come back to this in the future. This might be a feature that Sony is planning to add to Ci in the future, as I know that they were continuing development. I can let our contact know that this is something we would be interested in.

  *   No difference between video and audio embed. OK

  *   Only ORR will be embeddable: ie, if you're onsite and viewing something, we need to make sure that it's ORR before showing you the embed code. Correct

mccalluc:
@caseyedavis12 : So the merge brought in all the machinery, but it's hidden on the front end. At the redeploy we can make sure it really works, and then we'll make a tiny PR to make the code visible to users. We'll keep this open until that last bit is done.

PR #810, opened 2015-11-02
Embed player 799

mccalluc:
@afred: One question is how the embedded TOS should be styled. Right now I'm not pulling in any CSS so it gets the browser default serif, but it could be different.


mccalluc:
(Fix #808 + Fix #799)

afred:
:+1: looks good to me

PR #812, opened 2015-11-03
"_" in IDs is now handled by Cleaner.

### LINKED ISSUE #807: All ID cleaning in one place
Follow-up to #800: For that patch wanted to minimize changes to current output, but it will be better going forward if all ID-cleaning is done in one place, the `Cleaner`, and there's less tricky logic inside `PBCore`.

mccalluc:
Fix #807. @afred?

### DIFF
--- a/spec/scripts/downloader_spec.rb
+++ b/spec/scripts/downloader_spec.rb
@@ -29,8 +29,8 @@ describe Downloader, not_on_travis: true do
   it 'downloads by id' do
     dir = Downloader.download_to_directory_and_link(ids: ['cpb-aacip/17-00000qrv'], is_same_mount: true)
     expect(dir).to match(/\d{4}-\d{2}-\d{2}.*_by_ids_1/)
-    files = Dir["#{dir}/*.pbcore"]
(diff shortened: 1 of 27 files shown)


spec/scripts/downloader_spec.rb
```diff
@@ -29,8 +29,8 @@
   it 'downloads by id' do
```
mccalluc: This has actually been broken since we started zipping downloads, though we haven't noticed because this whole file is skipped in travis. Anyway, works now.

afred: @mccalluc good catch. Worth enabling in Travis?

mccalluc: @afred: No huge objections. I guess I had excluded this back when because
- I wanted to minimize test dependencies on outside services, particularly those that aren't required by the live site.
- and because if there have been a large number of recent updates in the AMS, the test can take some time.
File if you think it's justified and I'll do it.

afred: opened #814

PR #836, opened 2015-11-05
No more exhibit reference per item in results. Fix #831.

### LINKED ISSUE #831: Exhibits list view: Title Appears Twice
In Alan's exhibit, we need to include the Title and the URL in each section because he has organized his exhibit in bullet lists. Please see: http://americanarchive.org/exhibits/civil-rights/places

But when one clicks "show all items in this exhibit," the title that we added in the section markdown is added to the search results in the list view. We would like for this to not be shown in the list view.
screen shot 2015-11-05 at 12 25 26 pm

mccalluc:
@afred?


afred:
:+1: looks good, but also looks like some there were some tests checking for this. Go ahead and merge @mccalluc when tests are green.

PR #838, opened 2015-11-06
Really prefer British usage, but oh well. Fix #823.

### LINKED ISSUE #823: Fix graphic that appears on records in an exhibit
In the statement 

This record is featured in “Voices from the Southern Civil Rights Movement”.
 
the period at the end should go inside the quotation mark.

This needs to be fixed for all exhibits.

mccalluc:


afred:
I've never understood why the period goes inside. Makes no sense.

PR #842, opened 2015-11-06
Collapse orgs facet, even when selected.

### LINKED ISSUE #816: Clicking "View All Records" for an organization
From Alan: When you click on “View All Records” for organizations, the left side menu has the “organizations” tab open with all the other organization listed. When you click on one, you get a list that adds the titles from that organization to the one originally selected. I can’t figure out the use of that. It only confuses things. If you don’t click on an organization and close that tab, then click on another limiting term, the organization tab opens again.

mccalluc:
yay(?) monkey-patching. @afred? I'm not super confident about our diagnosis of the underlying problem, or of this UI hack to fix it, but it's not unreasonable either. I'm pretty confident that this is not something BL gives us out of the box, judging by the docs.
```
  # Determine whether a facet should be rendered as collapsed or not.
  #   - if the facet is 'active', don't collapse
  #   - if the facet is configured to collapse (the default), collapse
  #   - if the facet is configured not to collapse, don't collapse
```

Fix #816.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -74,7 +74,8 @@ class CatalogController < ApplicationController
     config.add_facet_field 'organization', sort: 'index', solr_params: { 'facet.limit' => -1 },
                            # Default is 100, but we have more orgs than that. -1 means no limit.
                                            tag: 'org', ex: 'org,state',
-                                           partial: 'organization_facet'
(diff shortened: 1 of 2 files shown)


afred:
@mccalluc I had [an idea about monkey patches](https://gist.github.com/afred/fec9b979d35f1424494b). Is this a good idea?

this PR looks good for now regardless. Merging.

Issue #820, opened 2015-11-04
Capitalization of titles


caseyedavis12:
From Alan: Capitalization of titles is not uniform. For example, the Pacifica titles are capitalized according to library rules (e.g., Asian American women vs. the women's movement), while others are not (e.g., Art and Violence).

Not sure if this can be done programmatically or if it needs to be done through manual cataloging.

mccalluc:
We're doing a tiny bit of capitalization cleaning [in code](https://github.com/WGBH/AAPB2/blob/master/scripts/lib/cleaner.rb#L228), and more could be done, but it only happens on a case by case basis.

In terms of things that are worth human time, tweaking capitalization ranks way, way down there for me, but that's just my opinion.

@caseyedavis12 : Can you narrow this down to something actionable, or close?

caseyedavis12:
@mccalluc: would it be difficult to implement capitalizing the first letter of every word in a title except for certain articles and other short words? I could give you a list of words to leave out. 

Reminder: please don't work on this yet until Karen and Nick confirm that you can work on this.

caseyedavis12:
Please proceed.

afred:
@caseyedavis12 as @mccalluc points out, if the title data is cased in a strange way, it's impossible to come up with rules that can guess correctly all the time, and in some cases, trying to be too clever will cause otherwise good titles to become oddly cased.

For one, we can't account for every acronym, so in some cases, we'd end up doing the wrong thing
  * "Hamers to be CEO of ING" would become "Hamers to be Ceo of Ing"
  * "WHO declares bacon deadly" would become "Who Declares Bacon Deadly"
  * "WILL will participate in AAPB" would become "Will Will Participate in Aapb"

But in other cases, some of the acronyms we do recognize may not be intended to be acronyms.
  * "DELIVER US FROM EVIL" would become "Deliver US From Evil" because we recognize "US" as acronym for United States. But, if we stopped recognizing "US" then...
  * "Terrorists Target US" would become "Terrorists Target Us".

It's very hard to say whether our rules would actually result in more correctness.

I suggest we do the following:
  * If the title is ALL CAPS _and_ more than 1 word, then go ahead and title case it, because even if it's incorrectly title-cased, it will look better than all caps.
  * Otherwise, use title as-is, because we just don't know if we'd be doing good or ill.

(NOTE: this is prime example of what we means when we say "we're playing whack-a-mole")
![image](http://i.imgur.com/moo0BtB.gif)

[...]

Issue #819, opened 2015-11-04
Articles in Titles


caseyedavis12:
Alan: Titles like “First Amendment, The” should appear as “The First Amendment” and can be sorted correctly if a non-displaying sort title field is created and populated with the correct way to sort titles.

mccalluc:
- see also #818 (sort by title)
- How common is this? I'm reluctant to invest much time on this unless we actually have a fair number of docs that would be affected.
- Also do "A"/"An"? Foreign languages?
- Implementation: probably just at render time.

PR #844, opened 2015-11-06
Exhibit title on search page.

### LINKED ISSUE #827: Title of Exhibit when showing all Items in exhibit
@mccalluc: When you click on “Show All Items,” the abstract displays like this:
 
Voices from the Southern Civil Rights Movement “Voices from the Southern Civil Rights Movement” presents educational and noncommercial radio programs from the 1950s and 1960s that offer historic testimonies – in interviews, speeches, and on-the-spot news reports – from many movement participants, both well-known and unknown. National leaders, local leaders, community organizers, students, clergy, lawyers, educators, academics, writers, and even a comedian and a documentary filmmaker relate often riveting stories that document a range of individual and group experiences and perspectives. The exhibit presents accounts from a variety of locales, each a distinct piece of the complex history of the struggle to integrate the segregated South and achieve full citizenship rights for African Americans.

It looks that the hyperlinked title is part of the first sentence of the text. It doesn’t look right. Can we have the title on its own line about the text, perhaps in the same way that the top of the narrative page looks – with the title above in a light blue band and maybe the “Exhibits” hyperlink available as well?

mccalluc:
Fix #827. @afred?

### DIFF
--- a/app/views/exhibits/_show_all_items_header.html.erb
+++ b/app/views/exhibits/_show_all_items_header.html.erb
@@ -1,2 +1,3 @@
-<%= link_to(exhibit.title, "/exhibits/#{exhibit.path}") %>
+<h4><%= link_to(exhibit.title, "/exhibits/#{exhibit.path}") %></h4>
+
 <%= sanitize(exhibit.summary_html, tags: []) %>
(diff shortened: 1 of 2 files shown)


mccalluc:
#819, too. I'm not totally convinced that cleaning every doc title is a good idea. Would consider moving this into the render code if we don't trust it, though I don't like that either.

mccalluc:
Made a new PR #853 for the good part, and we decided the capitalization was just a bad idea and killed #820.

PR #846, opened 2015-11-09
adds title lists and links to them from main page

sroosa:


sroosa:
I just realized that the links at the bottom of the main page don't work. Oops. Will fix now.

sroosa:
made some last minutes fixes. should be good to go, i believe.

afred:
@sroosa it should be `<br />` and not `</br>` for your HTML breaks

sroosa:
@afred Whoops! good catch! All fixed now.

PR #847, opened 2015-11-09
Sorting again

### LINKED ISSUE #818: Sort by Title
From Alan: It should ignore quotation marks at the beginning of titles. You may need to create a non-displaying “sort title” field, in which opening quotation marks are stripped out.

mccalluc:
@afred? Towards #818: Individual records will need to be reindexed before they sort correctly, but no new solr fields are created, so this doesn't require a reindex-all.

### DIFF
--- a/solr_conf/schema.xml
+++ b/solr_conf/schema.xml
@@ -464,7 +464,9 @@
             <filter class="solr.LowerCaseFilterFactory"/>
             <filter class="solr.TrimFilterFactory"/>
             <filter class="solr.PatternReplaceFilterFactory"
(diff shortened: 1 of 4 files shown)


afred:
@mccalluc good work! As a follow-up, do we want to come up with a solr query that will return records whose titles start with 'a', 'an', or 'the'?

mccalluc:
> come up with a solr query that will return records whose titles start with 'a', 'an', or 'the'?

- Not possible: They are stop words and so solr does not index them
- Articles were already working: it's the junk characters at the start of the string (also unindexed) that were the problem here.
- I **can** write a script to iterate over the index and find examples of those. Stand by...

PR #860, opened 2015-11-10
Fix broken affixed navigation.

### LINKED ISSUE #611: Floating Navigation on homepage
@Hammicakes: we have deployed the floating navigation to demo.americanarchive.org.

When I first come to the website, the floating navigation works great. But if I go to another page on the website and then come back to the homepage, the floating navigation does not appear.

afred:
* Adds jquery-turbolinks gem, which, resolves issues with jquery events not
  firing due to turbolinks black magic.
* Requires changing the event to which Affix plugin is applied.
* Closes #611.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,6 +22,10 @@ gem 'therubyracer', :platforms => :ruby
 gem 'jquery-rails'
 # Turbolinks makes following links in your web application faster. Read more: https://github.com/rails/turbolinks
 gem 'turbolinks'
+
+# A package that helps fix jquery code that breaks due to turbolinks.
+gem 'jquery-turbolinks'
+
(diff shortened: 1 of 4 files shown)


afred:
@mccalluc?

mccalluc:
@afred: This needs to be rebased on top of your other header PR in order to see it in action, right? I'm not seeing it right now. (Do we want to go on with this now, or would we like to experiment with turning it entirely off first?)

afred:
@mccalluc I don't think it needs rebased. You _do_ need to restart rails though

afred:
@mccalluc ok.. I'm now seeing what I assume to be the same brokenness. Rebasing, stand by.

afred:
@mccalluc? rebased. Looking good on my end.

PR #869, opened 2015-11-11
Reverts ID conversion

### LINKED ISSUE #868: Revert commit that moves ID translation logic from PBCore model into the Cleaner

afred:
Puts it back into PBCore model, instead of Cleaner.
Needed to fix ID chaos.
Closes #868


afred:
@mccalluc? check my adjustments to the tests

PR #882, opened 2016-01-14
Add blog to header

caseyedavis12:
@mccalluc: Can you make sure this looks OK?


mccalluc:
@caseyedavis12 : two problems:
- It looks like you're starting from an out of date version: If I apply the same change to the current master, tests pass.
- ... but, there are UI problems: In the medium width we're now getting a line wrap in the head, and that in turn screws up the display down below. See screen shot.
screen shot 2016-01-19 at 2 39 29 pm

We can't just keep adding links to the header indefinitely.

Your call on how to address this: you can propose an arrangement which preserves real estate, or we can live with the clutter and try to sort out the CSS. This PR is dead, though.

PR #884, opened 2016-01-14
Updating NET pages

sroosa:


afred:
if we want a 2nd pair of eyes on the content of this PR, I'm not sure @mccalluc nor I are the ones to do it.

Otherwise, if we just want to trust that @sroosa got it close enough, then we can merge it, only looking at passing tests.

Also, this is a lot of content (data?) to be managing with the CMLess stuff. Are we sure this is how we want to move forward?

Thoughts?

sroosa:
We decided (Casey and me and Karen) that we wanted to move forward like this, only because these pages are temporary. Like they will be up for a year or so as reference, as we work to get the catalog of NET titles where it needs to be in terms of the grant. Then we'll be ingesting it into the AMS, where we'll handle in like other records (with a possible few tweaks, which we can talk about during that phase of the project).

mccalluc:
@sroosa 
- What do you imagine happening when these pages do go away? Are you going to want something besides dead links?
- How do folks navigate down to these pages? Is there a list of the lists?

I'm not absolutely opposed to this, but if the goal is just to make these lists available, would you consider just making a big csv available for download? (Having it all in one place can really be a service to some users: If all I have is the title, easier to ctrl-f once, rather than once per page.) As a library I think we should try our best to avoid causing linkrot down the road.

mccalluc:
Looked at it a little more closely, and I found the navigation, but I remain worried about how useful this is for someone with just a title. I'm also worried about the data processing that produced this, and that makes me worried about the data processing going forward.

sroosa:
According to some examples of our user group (historians and people at stations) this is useful to them. In terms of the data processing, it was not very heavy because I still had the records at a spread sheet (we're working on getting them into Kevin's new pbcore database), from which I then selected the relevant columns and concatenate them to make each of these lines. After we post these updated pages, I think the plan is to leave them as is (not continually update them) and then delete the pages and possibly redirect all of the links to the NET Project page which will indicate that all of the NET titles are now represented in the AAPB records and can be searched for there. At least that's my thought right now.

afred:
Ok, so just to be clear, this is the plan then?....
  1. publish static NET pages (when this PR is deployed)
  1. get NET records eventually into a database (@foo4thought's FMP db)
  1. ingest NET records once we have a PBCore export
  1. edit verbiage on main NET project page to say "it's all online now, please search"
  1. add custom redirects for all static NET pages to go to the main NET project page
  1. delete static NET pages

sroosa:
That's the plan, except that between getting NET records into a database, and then exporting pbcore so that we can ingesting them into the AMS, we're going to be doing a lot of collaborative cataloging with the Library of Congress.

afred:
Super. merging.

PR #886, opened 2016-01-19
Fix tests

mccalluc:
@afred: I reverted @caseyedavis12 exhibit work since we don't want half-baked stuff in master, and I made a branch from this branch for her to work it. (The BBC is unrelated, but was failing for me.)


afred:
@mccalluc well, the half-baked stuff will forever be in master, unless we rewrite history with a `git reset` and a force push. We are very unlikely to accidentally deploy the broken build though, so I'm ok with leaving it in there in this case.

afred:
None of @caseyedavis12 commits added the test that this commit removes, and I don't see how any of her changes would have caused a test failure.

If that's the case, and the test failure is simply due to the BBC url suddently not being geolocated appropriately, then we don't need to revert anything. Just a forward patch that removes the test, right?

PR #890, opened 2016-01-19
Exhibition links

sroosa:
@mccalluc I think this fixes the problem you were talking about. Can you verify? Thanks!


mccalluc:
Sorry: You either need the full path (`/exhibits/station-histories/first-broadcasts`), including the first slash, or no slash at all (`first-broadcasts`), if you're coming from a page on the same level, but it can be kind of fragile.

sroosa:
@mccalluc - right. I always get that super mixed up. I will just go with the full path since it's easier for me to remember and less fragile.

PR #876, opened 2015-11-17
Bullets align better, and the banner is still at the top, but ...

mccalluc:
but not sure this is a good approach. Fix #874, perhaps. @afred?

PR #891, opened 2016-01-19
Fix bullet points, and nothing else in the header seems to change.

### LINKED ISSUE #874: Bullet points on /projects/permanent-entity
The bullet points on the /projects/permanent-entity page are not in alignment with the beginning of each statement in the list.

Done when: bullet points are aligned with the beginning of each statement in the list.

mccalluc:
Fix #874. I'm not seeing the problem we had on #876. @afred?

### DIFF
--- a/app/assets/stylesheets/override.scss
+++ b/app/assets/stylesheets/override.scss
@@ -4,7 +4,6 @@
   p:first-child {
     // Special handling for the banner image.
     margin: 0px;
-    display: inline-block;
     
     &:after {
       content: "";


afred:
@mccalluc fwiw, the `p:first-child` rule applies to many different elements through the override pages. However, I think it was originally only supposed to apply to the first _immediate_ descendent of `id="override-main"`. In any event, it looks like it's only effect (other than messing with the bullets) is a slight increased amount of spacing between header images and the first block of text. But if we really want that, then I'd rather go about it a different way anyway.

Also.. worth noting that other bullet points on override pages were not affected, because they were not wrapped in `<p>` tags. Not sure why, but we should probably try to ensure that CMLess markdown renders consistently in most cases.

mccalluc:
@afred: Can you articulate what you still want to see and file a bug for it?

afred:
@mccalluc looked a little deeper -- there's no bug

But FYI, @sroosa @caseyedavis12, when editing markdown files for AAPB, make not of how you create lists.

```markdown
- the items in this list
- will not be wrapped
- in <p> tags

However...

- the items in this list WILL be wrapped

- in <p> tags because of the extra

- line between the list items
```

The effect of this behavior is these two lists may display differently. If you don't want any spacing between your bulleted (or numbered) list items, then use the first method. If you _do_ want spacing in between your bulleted (or numbered) list items, then let us know, and we'll make sure the 2nd method gets that done for you.

PR #892, opened 2016-01-19
Move the default menu items into the partial itself

### LINKED ISSUE #872: "About" Menu on Exhibits Page
The "About" Menu on the Exhibits Page does not have the list of Projects in the Menu. 

Done when: List of projects appears below Projects on the About menu.

mccalluc:
since two separate instances need to show the same menu.

@afred: This fixes #872.

### DIFF
--- a/app/views/shared/_sidebar_nav.erb
+++ b/app/views/shared/_sidebar_nav.erb
@@ -1,36 +1,55 @@
-<% if !links.empty? %>
-  <nav class="navbar navbar-default">
-    <div class="container-fluid">
-      <!-- Brand and toggle get grouped for better mobile display -->
-      <div class="navbar-header">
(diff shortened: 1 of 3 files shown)


app/views/shared/_sidebar_nav.erb
```diff
@@ -1,36 +1,55 @@
...
+    ]],
+    ['Staff', '/about-the-american-archive/staff'],
+    ['Funding', '/about-the-american-archive/funding'],
+    ['Executive Advisory Council', '/about-the-american-archive/executive-advisory-council'],
+    ['Acknowledgments', '/about-the-american-archive/acknowledgments'],
+    ['FAQ', '/faq'],
+    ['Contact Us', '/contact-us']
+  ]
+%>
+<nav class="navbar navbar-default">
```
mccalluc: Only a whitespace change below this, since it's no longer wrapped in an 'if'.

PR #893, opened 2016-01-20
Redacted media

mccalluc:
@afred: This isn't complete, but how does this approach seem to you? Basically, the ci_ids, and eventually the media_srcs, will carry an extra bit of metadata about the visibility of that particular Ci asset. It keeps the surface interfaces the same at the cost of more complicated objects being passed underneath.

Alternatively, we could make pairs of methods where now we have only one:
`ci_ids` -> `protected_ci_ids` and `unprotected_ci_ids`
`media_srcs` -> `protected_media_srcs` and `unprotected_media_srcs`

I'm leaning towards the latter approach, now, but not really sure...


mccalluc:
@afred: Did some thinking last evening, and this is not the right approach. Look for another PR latter today.

PR #897, opened 2016-02-01
Presidential elections

caseyedavis12:
We're ready to merge and deploy. @sroosa has the list of GUIDs to reindex.


afred:
I'm adding a bit to even out the carousel, so we can get all of them in there, while not leaving empty spaces. Stand by...

afred:
@mccalluc?

mccalluc:
@afred: I'll try to fix the test and will review the substance.

afred:
the PR is intended to avoid this situation...

image

mccalluc:
@caseyedavis12 : footnotes on http://localhost:3000/exhibits/presidential-elections/candidates point to http://localhost:3000/exhibits/presidential-elections/notes#7, but there doesn't seem to actually be a matching anchor on that page?

caseyedavis12:
@mccalluc: Does the anchor on the notes page need to look like


```<a name="7"></a><sup>7</sup>```


or


```<a name="#7"></a><sup>7</sup>```

mccalluc:
@caseyedavis12 : the former: `<a name="7"></a><sup>7</sup>`

The "#" is punctuation in the URL. Also, that will will, but if anything else on the page happens to have an id that's just a number, your links will break. Safer, as we discussed, to use some consistent prefix in your anchors. (But again, that's just future-proofing.)

mccalluc:
The notes page isn't working because the content hasn't been put in a section, and so it ends up inheriting content from the top page: ie, there's content on the page, just not the content we want. I'll fix. @caseyedavis12 : this explains the broken links.

afred:
@mccalluc the broken tests are due to `api.js` returning 1 fewer results that expected. However, looking at the tests, it's unclear to me whether this is really an introduced bug, or whether the tests just need to be adjusted.

afred:
@mccalluc hmm.. so now I'm confused as to what was causing [this failure](https://travis-ci.org/WGBH/AAPB2/builds/106259299)?

caseyedavis12:
@mccalluc: Does content in exhibit sections have to be in the Online Reading Room


There is one program that is not yet in the online reading room that is linked in an exhibit section. I am waiting on approval. I can remove it if that will fix the issue.

caseyedavis12:
@mccalluc: This is the program for which I am waiting approval:
cpb-aacip/15-st7dr2pn78

afred:
Ugh... failures due to a connection timeout error w/ Solr. Trying again.

afred:
passing! [see this](https://travis-ci.org/WGBH/AAPB2/builds/106267698)

merging!!

PR #914, opened 2016-02-24
Update _access_facet.erb

caseyedavis12:
Editing the messaging about content that is not digitized. Gives users the wrong impression that we are currently continuing to digitize.

### DIFF
--- a/app/views/catalog/_access_facet.erb
+++ b/app/views/catalog/_access_facet.erb
@@ -17,7 +17,7 @@
       {
         value: PBCore::ALL_ACCESS,
         title: 'All Records',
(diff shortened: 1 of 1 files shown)


mccalluc:
Not sure what's up with the failures, but I'll look into it. 

@caseyedavis12 : Do you also want to change the messaging around 'on-site-only', since we're now allowing scholars anywhere to have access?

caseyedavis12:
@mccalluc: We're still waiting for legal to finalize the language for the password-protected access policy, unfortunately, so we need to hold off on changing facet messaging.

mccalluc:
just rebased and pushed...

PR #919, opened 2016-04-08
Explicit extensions are good.

mccalluc:
@afred?


mccalluc:
(I'll rebase...)

mccalluc:
(File renames plus edits are causing confusion, I think. Will do this over.)

PR #927, opened 2016-04-11
sticky sidebar. Fixes #712.

### LINKED ISSUE #712: About page floating nav
grab from mike's branch, or do it yourself.

mccalluc:
@afred?

### DIFF
--- a/app/assets/javascripts/sticky.js
+++ b/app/assets/javascripts/sticky.js
@@ -1,0 +1,9 @@
+$(function(){
+    var $can_stick = $('.can-stick');
+    $can_stick.width($('.panel-primary').width());
+    var offset = $can_stick.offset()['top'];
+    $(window).scroll(function(){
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/sticky.js
```diff
@@ -0,0 +1,9 @@
+$(function(){
+    var $can_stick = $('.can-stick');
+    $can_stick.width($('.panel-primary').width());
```
mccalluc: So things may get wonky with window resizes, but it works around some other problems...

afred:
@mccalluc code looks ok to me. Can we see a screen shot of the various states?

mccalluc:
Here are two questionable conditions:
- If the window shrinks, it keeps the old width.
- Floats under the footer.

screen shot 2016-04-11 at 11 59 38 am

PR #929, opened 2016-04-11
We have not really been using coverage.

mccalluc:
@afred? I think this could have been useful if
- we were trying to get to 100% coverage
- and we were connected to coveralls, or something like that.

The report generation adds a few seconds, and if we're never looking at it, it's a waste of time.


afred:
m'ok. I would support an effort to increase test coverage, but we can make new tickets for that when we're motivated to do so.

PR #936, opened 2016-04-12
Better link checker

mccalluc:
@afred?


mccalluc:
huh. I am not reproducing this failure locally.

mccalluc:
@afred, working now, if it looks good to you.

spec/support/link-check-ignore.txt
```diff
@@ -0,0 +1,165 @@
...
+https://s3.amazonaws.com/americanarchive.org/resources/pbcore_excel_template.xls
+https://twitter.com/amarchivepub
+https://weru.org/
+https://www.congress.gov/bill/112th-congress/house-bill/1473/text
+https://www.facebook.com/amarchivepub
+https://www.law.cornell.edu/uscode/text/2/170
+https://www.nedcc.org/free-resources/dplan-the-online-disaster-planning-tool
+https://www.sonymcs.com/
+https://www.youtube.com/watch?v=dv2JEsHiL8U
+https://wwws.loc.gov/readerreg/remote/
```
afred: @mccalluc why is the ignore list so long? Don't we want to be notified if one of these links dies?

mccalluc: It would be nice, but it's not a check that should be made with every travis run. In the earlier version I had this tricky only-check-when-run-locally-and-only-when-it-hasn't-been-run-in-the-last-month logic, that was never that useful. 

At any point that we choose we can move this file temporarily out of the way and run the tests.

PR #938, opened 2016-04-14
Error capture the same as OV. Fix #905.

### LINKED ISSUE #905: More consistent error handling on production
Going to http://americanarchive.org/catalog/foo results in this...
image

Going to http://americanarchive.org/catalog/foo/bar results in this...
image

And going to http://americanarchive.org/catalog/cpb-aacip_394-515mkz2f/foo results in this...
http://americanarchive.org/catalog/cpb-aacip_394-515mkz2f/foo

The last example is what all invalid routes should point to.

**Done when**
  * invalid routes do not result in a bubbled up 500 error
  * invalid routes display the custom 404, and not the rails default 404

mccalluc:


mccalluc:
better to start from scratch

Issue #944, opened 2016-04-20
Popup archive getting 401 unauthorized when trying to access media URLs


afred:
**background:** [In this code](https://github.com/WGBH/AAPB2/blob/3574dcb82e1870aefb79122bfe8899b86e3a9a07/app/controllers/media_controller.rb#L9-L15) we are hitting the `else` clause which responds with the "401 Unauthorized". The reason we're falling into the `else` is because the the `can?(:play, pbcore)` condition is failing, and that is failing because of [how it is defined](https://github.com/WGBH/AAPB2/blob/3574dcb82e1870aefb79122bfe8899b86e3a9a07/app/models/ability.rb#L7-L10)... in particular, because `popuparchive.org` is not considered "on site", it's requiring the check for TOS, which is failing.

This could be solved by just adding `popuparchive.org` to the IP ranges considered to be "on site". If we don't want to do that, then we need to adjust the conditional logic to make an allowance for the `popuparchive.org` referrer.

**Done when** media URLs do not return 401 when referrer is `popuparchive.com`

afred:
@mccalluc this cropped up while you were gone. I got as far as PR #945, but I'm not sure it solves anything, because I've been having a hard time reproducing their error.

Any ideas on where things might be getting hung up?

mccalluc:
Two lessons learned:
- `--referer 'http://americanarchive.org/;auto`
- DO NOT use `-I`: S3 does not accept `HEAD` requests, though it does accept `GET`.

mccalluc:
@afred / @caseyedavis12 : I think we determined that it's not a problem with our code, but rather with the curl arguments they were using. I wasn't on the thread... can one of you follow up with them and close this is nothing else needs to be done on our end?

caseyedavis12:
@afred would you be willing to send a note to Bailey and Anne confirming that we have done all we can do regarding this issue?

________________________________

afred:
you bet. Stand by.

afred:
followed up with email to Bailey and Anne, instructing to try the modified curl command, with the correct value for `--referer` and without the `-I` option, and if it still doesn't work, then we should set aside some time to do live trouble shooting.

closing.

PR #945, opened 2016-04-21
Fixes specs for media urls

afred:
Fixes a bug in media controller.
Refactors spec for tesing media url redirection.
Adds a test for media url redirection given referer of popuparchive.com.
Causes User#aapb_referer? to return false in case of missing or malformed referer value.

### DIFF
--- a/app/controllers/media_controller.rb
+++ b/app/controllers/media_controller.rb
@@ -9,7 +9,7 @@ class MediaController < ApplicationController
     if can?(:play, pbcore) && (current_user.aapb_referer? || current_user.embed?)
       ci = SonyCiBasic.new(credentials_path: Rails.root + 'config/ci.yml')
       # OAuth credentials expire: otherwise it would make sense to cache this instance.
-      redirect_to ci.download(pbcore.ci_ids[(params['part'].to_i || 1) - 1])
+      redirect_to ci.download(pbcore.ci_ids[(params['part'] || 1).to_i - 1])
     else
       render nothing: true, status: :unauthorized
     end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -3,50 +3,62 @@ require 'ipaddr'
 require_relative '../../lib/geo_i_p_country'
 
 class User
+  POPUP_HOST_RE = /^(.+\.)?popuparchive\.com$/
+  AAPB_HOST_RE = /^(.+\.)?americanarchive\.org$/
+  WGBH_IP_RANGE = IPAddr.new('198.147.175.0/24')
+  LOC_IP_RANGE = IPAddr.new('140.147.0.0/16')
+
+  attr_reader :request
+
   def initialize(request)
-    @ua = request.user_agent
-    @ip = request.remote_ip
-    @referer = request.referer
-    @session = request.session
+    @request = request
   end
 
-  ONSITE_RANGES = Set[
-    IPAddr.new('198.147.175.0/24'), # WGBH
-    IPAddr.new('140.147.0.0/16') # LoC
-  ] + (Rails.env.production? ? [] : [IPAddr.new('127.0.0.1')])
-
   def onsite?
-    ONSITE_RANGES.map { |range| range.include?(@ip) }.any?
+    onsite_ip_ranges.map { |range| range.include?(request.remote_ip) }.any?
   end
 
   def usa?
-    GeoIPCountry.instance.country_code(@ip) == 'US' || onsite?
+    ::GeoIPCountry.instance.country_code(request.remote_ip) == 'US' || onsite?
     # WGBH doesn't actually geocode to USA. No idea why.
   end
 
   def bot?
-    /bot|spider/i =~ @ua
+    !(/bot|spider/i =~ request.user_agent).nil?
+  end
+
+  def referer_host
+    URI.parse(request.referer).host
+  rescue URI::InvalidURIError
+    nil
   end
 
   def aapb_referer?
-    [
-      /^(.+\.)?americanarchive\.org$/,
-      POPUP_HOST_RE,
-      /^54\.198\.43\.192$/
-    ].any? do |allowed|
-      URI.parse(@referer).host =~ allowed
+    aapb_referer_regexes.any? do |aapb_referer_regex|
+      referer_host =~ aapb_referer_regex
     end
   end
 
   def embed?
-    URI.parse(@referer).path =~ /embed/
+    URI.parse(request.referer).path =~ /embed/
   end
 
   def affirmed_tos?
-    @session[:affirm_terms] ||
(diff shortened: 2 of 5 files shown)


afred:
@mccalluc?

This is toward #944, but I'm not sure it actually changes any behavior with respect to how requests are being handled coming from popuparchive.org.

I may try pushing this branch to the demo server and testing it with curl

afred:
and when I say "fixes specs" in the PR description, I don't mean that specs were broken, and now they pass. Rather, the specs for testing access to media urls based on referer checks were returning false positives because localhost was considered an "on site" IP address, so the condition we really wanted to test was not being tested.

app/controllers/media_controller.rb
```diff
@@ -9,7 +9,7 @@ def show
     if can?(:play, pbcore) && (current_user.aapb_referer? || current_user.embed?)
       ci = SonyCiBasic.new(credentials_path: Rails.root + 'config/ci.yml')
       # OAuth credentials expire: otherwise it would make sense to cache this instance.
-      redirect_to ci.download(pbcore.ci_ids[(params['part'].to_i || 1) - 1])
+      redirect_to ci.download(pbcore.ci_ids[(params['part'] || 1).to_i - 1])
```
mccalluc: What's the bug, and what's the desired behavior?

app/models/user.rb
```diff
@@ -3,50 +3,62 @@
...
+  private
+
+  def aapb_referer_regexes
+    [AAPB_HOST_RE, POPUP_HOST_RE]
+  end
+
+  def onsite_ip_ranges
+    @onsite_ip_ranges ||= begin
+      ranges = [WGBH_IP_RANGE, LOC_IP_RANGE]
+      ranges << IPAddr.new('127.0.0.1') if Rails.env.development?
```
mccalluc: What would be the harm of always including this? Env conditionals scare me a little.

app/models/user.rb
```diff
@@ -3,50 +3,62 @@
 require_relative '../../lib/geo_i_p_country'
 
 class User
+  POPUP_HOST_RE = /^(.+\.)?popuparchive\.com$/
+  AAPB_HOST_RE = /^(.+\.)?americanarchive\.org$/
+  WGBH_IP_RANGE = IPAddr.new('198.147.175.0/24')
+  LOC_IP_RANGE = IPAddr.new('140.147.0.0/16')
+
+  attr_reader :request
```
mccalluc: Does anything outside actually need to access this? If not, I'd prefer not to expose it, and internal use can be `@request`

app/models/user.rb
```diff
@@ -3,50 +3,62 @@
...
 
   def usa?
-    GeoIPCountry.instance.country_code(@ip) == 'US' || onsite?
+    ::GeoIPCountry.instance.country_code(request.remote_ip) == 'US' || onsite?
     # WGBH doesn't actually geocode to USA. No idea why.
   end
 
   def bot?
-    /bot|spider/i =~ @ua
+    !(/bot|spider/i =~ request.user_agent).nil?
```
mccalluc: just to return an actual `true` instead of a true-ish integer? Is there another problem?

app/models/user.rb
```diff
@@ -3,50 +3,62 @@
...
+  end
+
+  def referer_host
+    URI.parse(request.referer).host
+  rescue URI::InvalidURIError
+    nil
```
mccalluc: Maybe `''` would be better, and then you could use the `||=` idiom? This is going inside a loop, albeit a small one.

app/models/user.rb
```diff
@@ -3,50 +3,62 @@
...
-    URI.parse(@referer).path =~ /embed/
+    URI.parse(request.referer).path =~ /embed/
   end
 
   def affirmed_tos?
-    @session[:affirm_terms] ||
-      (!@referer.nil? && URI.parse(@referer).host =~ POPUP_HOST_RE)
-    # Casey confirms that Popup counts as affirming ToS.
+    # Referer's from popuparchive.com get a pass on TOS
+    request.session[:affirm_terms] || !(referer_host =~ POPUP_HOST_RE).nil?
```
mccalluc: Does anything go wrong if you don't cast the integer to boolean?

spec/features/catalog_spec.rb
```diff
@@ -329,6 +329,10 @@ def expect_external_reference
   end
 
   describe '#show' do
+    before do
+      page.driver.options[:headers] = { 'REMOTE_ADDR' => '198.147.175.1' }
```
mccalluc: add comment explaining that this IP is WGBH?

app/controllers/media_controller.rb
afred: when `params['part']` is `nil`, then...
* `(params['part'].to_i || 1) - 1`
* => `(nil.to_i || 1) - 1`
* => `(0 || 1) - 1`
* => `0 - 1`
* => `-1`

and, because it's ruby, passing a negative number as an array index means count backward from the end of the array.

app/models/user.rb
afred: nope, but i think sticking with the idiom that methods ending in `?` should `true` or `false` has value. It's confusing to me when...
```ruby
puts "it's a bot" if bot?
# => it's a bot
puts "it's not a bot" unless bot? == true
# => it's not a bot
```

app/models/user.rb
afred: should be able to use the `||=` idiom regardless of the return value...
```ruby
def referer_host
  @referer_host ||= begin
    URI.parse(request.referer).host
  rescue URI::InvalidURIError
    nil # or empty string if you prefer
  end
end
```

i'm down with that.

app/models/user.rb
afred: i don't think anything in our code goes wrong. But if you tried comparing the return value to `true` or `false` with `==`, then it would always return `false`, which is surprising to me.

app/models/user.rb
afred: This is just a [refactor from above](https://github.com/WGBH/AAPB2/pull/945/files/e6bc391bcb97626affa24ce7da97c9344c78aafe#diff-4676c008b11a5480d73d4a6de01e45b9L16).

The logical change is that before, it considered localhost as "on site" whenever you _weren't in production_. Now it considers localhost as "on site" _only when you are in development_.

Effectively, this says that when then ENV is "test", you are _not_ "on site".

The test I had put in for testing requests from popuparchive.org were getting false positives, because it was assuming that all requests were "on site". It was never getting to the condition that I really wanted to test.

There is a comment in `app/models/ability.rb` to uncomment out a condition when testing TOS features, but that's not a very consistent way to test, and it's not very flexible.

app/models/user.rb
mccalluc: Either way, though if you get the nil it will need to reevaluate anyway.

PR #947, opened 2016-04-26
adds wnet history

sroosa:


sroosa:
@mccalluc Will you check that I formatted it correctly in the .yml. I tried to copy what we did with the other stations history, but the text seems like it's wrapping weirdly. Thanks!

mccalluc:
Rebase might help with tests?

mccalluc:
@sroosa : I've loaded it locally, and the long lines cause no problems in display. I'll put the manual line wraps in there, but there's nothing special about it: just hit return.

mccalluc:
Remaining test failures after rebase are related to #949, and are unrelated to PR.

Issue #943, opened 2016-04-20
Update Ruby version


afred:
version 2.0.0 is end-of-life as of 2/24/16

caseyedavis12:
@afred: can this be closed?

afred:
Yep, we're on 2.2.2 now.. but actually, 2.3 has been out for a year, so probably time for another upgrade. Closing this in favor of #1193

PR #952, opened 2016-04-26
Update ruby version

mccalluc:
@caseyedavis12 : Keeping versions up to date is a good thing, but sometimes dependencies haven't kept up. Do we believe that gallery remains important? It might be a quick fix, or it might not, but if it's not essential, dropping it would be the easiest solution.

Towards #943.


caseyedavis12:
@mccalluc: Do you mean that with this ruby update, we lose the Gallery view on AAPB unless further steps are taken to re-implement? Would a curated exhibit section just use the list view below the narrative?

________________________________

mccalluc:
Casie is ok with dropping gallery.

afred:
@mccalluc what was the answer to @caseyedavis12's question?.. i.e .what does "dropping the gallery" imply?

afred:
@mccalluc there appear to be failures other than the SSL issue

mccalluc:
@afred: I'm a bit perturbed by the apparent change in the DOM structure that Capybara is now seeing: it seems to think there's something like an empty element, but when I inspect the html source, I don't see what it's seeing.

Any other thoughts?

Issue #946, opened 2016-04-25
Remove non-alaphabetic (non-word?) characters for title sort


mccalluc:
Follow up from #425. After code is deployed, either do a reindex, or just the first 12430 in this list: http://americanarchive.org/catalog?f[access_types][]=all&page=1243&sort=title+asc

@caseyedavis12 : Originally you specified that leading numbers be removed... I think it would be simpler and just as good to remove leading non-alphanumerics, but to sort all numbers to the end, keeping the existing z padding. So, in the end it should sort like:

- Cat Pictures
- A Catalog
- Zoom
- 3-2-1 Contact
- "28 Days Later"
- !@#$%^&* 101 Dalmations
- 700 Club
- !@#$%^&*

mccalluc:
Hold on... it should be doing that already; Will investigate.
```xml
            <filter class="solr.PatternReplaceFilterFactory"
                    pattern="[^\w\s]+" replacement="" replace="all"/>
```

mccalluc:
@caseyedavis12 / @afred : I am pretty confident that the problem is just that it needs to be reindexed.

I've checked that the analysis solr performs is what we want: http://localhost:8983/solr/#/blacklight-core/analysis?analysis.fieldvalue=@@@leading-punct-is-stripped&analysis.fieldtype=sort

I've ingested just a single record locally, and the behavior is correct: (local on left, production on right)

<img width="924" alt="screen shot 2016-04-27 at 3 34 27 pm" src="https://cloud.githubusercontent.com/assets/730388/14865682/bb65574c-0c8d-11e6-8d9c-f1dd9b821ef8.png">

I think the next step is just to collect a list of the 12K IDs that need to be reindexed, and then then on production `download_clean_ingest.rb --just-reindex --id-files ids.txt`.

mccalluc:
caveat: The character "_" does count as "\w", so the solr should be tweaked to be strictly alpha-numeric.

mccalluc:
Here's the script that does the job: 
```ruby
require 'open-uri'
require 'json'

0.step(1000000, 100) do |start|
  json = URI("http://americanarchive.org/api.json?fl=id,title&rows=100&start=#{start}&sort=title+asc").read
  JSON.parse(json)['response']['docs'].each do |doc|
    if doc['title'].sub(/^\s*(a |an |the )?/i, '') =~ /^[Aa]/
      warn "Found #{doc['title']}"
      exit
    end
    puts doc['id']
  end
[...]

PR #955, opened 2016-04-27
Fix one small bug, and add low level test. Towards #946.

mccalluc:
@afred?


mccalluc:
looks good to @foo4thought

PR #957, opened 2016-05-02
Update deploy docs

mccalluc:
I do not want to copy-and-paste everything from AAPB, but this is more accurate. Not sure about the future. One possibility is just keeping a brief outline here, and pointing at OV for the rest. Should there be a deploy.sh for this project as there is for OV?

@foo4thought ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,10 +49,15 @@ Wordpress sets a default TTL of 300s, which should be fine for now.
 
 ### Deploying to production
 
-Code and documentation are hosted in the private repository,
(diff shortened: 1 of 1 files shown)


foo4thought:
Redundacy of accurate documentation across projects is no vice.  I have no opinion about deploy.sh  except that if it fits, why not wear it?

mccalluc:
(Except that you run into the same kind of problems as when copy-and-pasting code?...)

afred:
👍 merging. Create a new ticket for `deploy.sh` if needed.

PR #959, opened 2016-05-04
Strengthen tests, ignore SSL everywhere.

mccalluc:
Will self merge after tests pass.


spec/scripts/downloader_spec.rb
```diff
@@ -2,7 +2,7 @@
 require_relative '../../scripts/lib/downloader'
 
 describe Downloader do
-  it 'can download the past 7 days', not_on_travis: true do
+  it 'can download the past 7 days' do
```
mccalluc: because this test was turned off on travis, we missed this bug. Tests will be a bit slower now, though we hopefully won't notice.

PR #960, opened 2016-05-05
Update header nav-bar

caseyedavis12:


caseyedavis12:
Doing QC on demo and noticed this needed to be updated to reflect the changes to the page. @mccalluc, can we include this in the redeploy?

PR #964, opened 2016-05-05
Style reorg

mccalluc:
@Muraszko : Back when, Mike H delivered us a huge, monolithic CSS file. This is getting to be unmaintainable. This PR moves us towards the pattern of OV, without actually changing any behavior (hopefully).

- The files which would have been covered by the `include_tree` are now listed individually under `globals/`, and they are included in alphabetical order, replicating the behavior of include_tree.
- The remaining code in `application` has been split into coherent sections, and these are included in the order they occurred in the original file.
- The file extensions have all been changed to `.css.scss`, though no scss functionality is in use.

I think this is low risk, but I also think it'd be good to do some QA before merging. If there are problems, it will be easier to spot them now, than latter.

At some point in the future:
- actually use scss
- check that code grouping makes sense, within files, and between local and global.


Muraszko:
@afred Sounds good to me, you definitely want to have @caseyedavis12 review since it's AAPB.  Thanks.

afred:
@mccalluc kludge?

afred:
👍 renaming looks fine to me.

mccalluc:
@afred :
I can't tell from your initial comment whether you have an underlying concern? It is kludgy, in the sense that it is not really SCSS... but I didn't want to go there in one leap, because if there are small regressions, they'll be easier to spot with small merges.

@caseyedavis12 :
I think folks want you to merge this. Let me know if you need more background on the motivation for this change?

afred:
@mccalluc no concern, was just wondering what the kludge was.

app/assets/stylesheets/locals/kludge.css.scss
```diff
@@ -0,0 +1,17 @@
...
+	.clearfix:after {
+		visibility: hidden;
+		display: block;
+		font-size: 0;
+		content: " ";
+		clear: both;
+		height: 0;
+	}
+* html .clearfix             { zoom: 1; } /* IE6 */
+*:first-child+html .clearfix { zoom: 1; } /* IE7 */
```
mccalluc: @afred: Sorry: I misunderstood you. As I was pulling things out, my first concern was to preserve order, in case one thing overrides another. For all the other files, there was some theme that applied, but there were just stray pieces left for this one. Maybe etc.css.scss would have been a better name.

PR #965, opened 2016-05-09
update newshour

caseyedavis12:
removed the image with the logos

@mccalluc?


mccalluc:
(It's stalling on the `Downloader` test, which we recently expanded the scope of... Wait a bit longer and see if it completes.)

PR #969, opened 2016-05-11
Update footer

caseyedavis12:


mccalluc:
(failure is the rubocop complaint on the unfrozen constant. I'll fix.)

mccalluc:
@afred: Unless you see a problem, I think this makes sense: use Rail's built-in caching rather than rolling our own.

PR #982, opened 2016-05-19
Oai thumbnail 981 plus merge

mccalluc:
@foo4thought ?

(Another part of the philosophy of code reviews, for me, is that even if it's trivial, it gives the other person a gradual introduction to a code base, and helps to make those connections between a vague problem and a specific solution.)


foo4thought:
MVC ethics question:  does something specific to implementation like 'http://americanarchive.org.s3.amazonaws.com/thumbnail/1234.jpg' better belong in a controller instead of model?

mccalluc:
@foo4thought : are you referring to `spec/fixtures/pbcore/clean-MOCK.mods`, or is there something else? That is neither an M or a V, but a part of the tests.

foo4thought:
D'oh.  It was in the test; I thought it was in model code.

PR #983, opened 2016-05-23
Transcripter

mccalluc:
@foo4thought or @Muraszko ? The idea is to take the SRTs we already have and render them as html so we can have an iframe like on OV. Partway there:

screen shot 2016-05-23 at 1 42 40 pm

(This will only show when you append `?transcript=yes` to the URL.) Right now, the player doesn't seem to be sending `timeupdate` events: It might be that the JS wrapper we're using is swallowing the events, and I think the easiest fix will just be to use plain html5 video.

The bigger picture: We don't know what format digital will deliver, but unless we need something more complicated, let's try to use SRT as the base, which will be translated into webvtt and html as needed. I'd prefer to avoid having multiple base formats, and we already have some SRT: that's the only reason I'm starting there.


mccalluc:
@foo4thought or @Muraszko ?

Issue #962, opened 2016-05-05
Make widths of elements on "about" pages consistent for medium and narrow widths


afred:
The varying widths of elements looks weird to me.
![image](https://cloud.githubusercontent.com/assets/1507490/15046606/b15a506e-12ae-11e6-8ffa-f4fc733b5050.png)

I think it would look better if all those elements were the same width.

afred:
p.s. this is just me pretending to be a designer. Feel free to ignore/close this if you disagree, or don't think it's important.

PR #985, opened 2016-05-23
Width weirdness

mccalluc:
Towards #962. The CSS is really a mess, and I'm not sure how much ad-hoc tweaking we want to get into.


mccalluc:
@foo4thought or @Muraszko ?

PR #986, opened 2016-05-25
CORS: fix #686. jquery.xmleditor now can pull AAPB records.

### LINKED ISSUE #686: CORS API?
If we want an API at some point, CORS in the right way to do it... but just dropping in https://github.com/cyu/rack-cors is not working for me. Tried a few permutations, but no success. In the Chrome console I get this message:
>
No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'null' is therefore not allowed access. The response had HTTP status code 400.

(The requesting page is just on `file:///`, so that's the null origin, and that might be the real problem? Maybe * doesn't match null?)

mccalluc:
@foo4thought ? If you haven't heard of CORS, I could try to explain the motivation.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -68,6 +68,7 @@ gem 'rubyzip'
 
 gem 'cancan'
 gem 'geoip'
+gem 'rack-cors', :require => 'rack/cors'
 
 gem 'libv8', '~> 3.16.14.13' # Chuck had problems with older version.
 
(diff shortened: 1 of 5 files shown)


mccalluc:
argh. These tests are working locally, but the OPTIONS requests when run on Travis produce 500s. There is more wiring (https://gist.github.com/dhoelzgen/cd7126b8652229d32eb4) but I'd like to reproduce the failures locally before just adding more code that I only half-understand.

mccalluc:
Start with `RAILS_ENV=test bundle exec rails server` to reproduce.

mccalluc:
@foo4thought or @Muraszko ?

Muraszko:
OK to go after conflict fix.

PR #988, opened 2016-06-06
Orgs yaml to cmless

mccalluc:
@sroosa : With the couple station edits you'll be doing, thought it'd be a good time to break it out of the yaml... Does this seem ok?

@Muraszko : code review and merge?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -47,7 +47,7 @@ gem 'capistrano-passenger', :group => :development
 # Use debugger
 # gem 'debugger', group: [:development, :test]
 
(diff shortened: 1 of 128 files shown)


sroosa:
This seems good to me! All I have to do is add a new md file for each new org, right?

mccalluc:
Right: the name of the file needs to match the internal ID.

sroosa:
@mccalluc : This looks good to me. Can I do the code review in merge, or does it have to be @Muraszko ?

PR #990, opened 2016-06-07
Fix #989 and test. 

### LINKED ISSUE #989: turn off thumbnail referer checks

mccalluc:
Security is less tight than OV, so this is a bit overkill. @Muraszko ?


mccalluc:
@Muraszko ?

PR #1010, opened 2016-07-07
Chrome transcript error goes away

### LINKED ISSUE #1005: Cross-site security for transcripts
Chrome has this helpful message on http://americanarchive.org/catalog/cpb-aacip_111-02c8693q :
>
Text track from origin 'https://s3.amazonaws.com' has been blocked from loading: Not at same origin as the document, and parent of track element does not have a 'crossorigin' attribute. Origin 'http://americanarchive.org' is therefore not allowed access.

(FF is not working, either, but there is no error message. When Chrome is fixed, see if that also helps FF.)

mccalluc:
but still no display of transcript.

Maybe VTT instead of SRT would fix that? Fix #1005. @foo4thought ?


foo4thought:
it's easy enough - so I'm told - to convert to VTT from SRT, so it's a reasonable test.

PR #1011, opened 2016-07-07
Remove --same-mount. Fix #1009.

### LINKED ISSUE #1009: Remove --same-mount

mccalluc:
@foo4thought : If we don't need --same-mount, then let's get rid of it.


foo4thought:
I have deployed code that uses the switch your about to deprecate; will it throw an error now or must I modify and redeploy?

mccalluc:
Not right now, but once #1018 is in, it will.

PR #1029, opened 2016-07-26
Update show.html.erb

Muraszko:
PR to put the embed link on every record page.  Hold off merging until @caseyedavis12 gives the ok.

### DIFF
--- a/app/views/catalog/show.html.erb
+++ b/app/views/catalog/show.html.erb
@@ -200,7 +200,7 @@
   <% end %>
       
   <% if @pbcore.public? %>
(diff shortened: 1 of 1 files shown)


afred:
@Muraszko 👍. Feel free to merge if/when tests pass.

This one you'll probably want to do some QC on demo before going live with it.

Muraszko:
@afred Thanks for getting these to pass.  Still holding off on this branch until @caseyedavis12 gives the ok.

Muraszko:
Update: Casey has said this is no longer a feature we want to provide. Closing.

PR #1034, opened 2016-08-15
Fixes CORS issue with video tracks

afred:
Changes value of `crossorigin` property from "anonymous" to "with-credentials",
which allows the caption file to be served from S3, and also sends the cookie
containing the session ID so that the MediaController can successfully retrieve
the user's session, which has the correct value for indicating the user has
agreed to the TOS, which allows the video to play.


foo4thought:
it looks like Travis stumbled when it tried to pull what seems like a good URL; will it re-run or do we have to kick it?

afred:
also failed because of a bunch of damn rubocop errors. Those have been fixed. Going to merge this and deploy to demo server to see what happens, yay!

PR #1050, opened 2016-09-21
Update catalog_controller.rb

Muraszko:


afred:
@Muraszko did you get the test to pass?

Muraszko:
No, but the only change was text.
When I removed the Ci script directory this morning (before I put it back) I was hoping that was what was throwing the tests off but it wasn't the case, it still failed:
https://travis-ci.org/WGBH/AAPB2/builds/161623824

afred:
The failing test is one that is checking for specific text. You change the text, you need to change the test as well.

This isn't the best way to test this type of thing, for exactly this reason. But I think we still want all the tests to be green. I'll take a quick peek.

Muraszko:
@afred Ahh, that makes sense. Yes, green tests would be good. Thank you!

PR #1051, opened 2016-09-21
Update README.md

Muraszko:
Updating the README to reflect the correct deployment repo so as to not keep documentation on the broken one.
@afred ?


afred:
@Muraszko 👍. Feel free to merge if/when tests pass.

PR #1052, opened 2016-09-21
Fixes test

afred:


afred:
@Muraszko merge?

afred:
after this is merged, I'll rebase the other two branches, and they should go green too.

PR #1057, opened 2016-09-25
Roll the logs based on size

afred:


afred:
@Muraszko review and merge?

PR #1060, opened 2016-09-28
Gets captions working

### LINKED ISSUE #1054: Closed captioning still not appearing for records with captions
I am sorry if this is duplicative of another issue but I could not find it. Closed caption files are still not appearing with the player for records that have an SRT file stored on AWS. I have tried in Chrome and Firefox and the issue is present in both.

afred:
Involves a non-trivial refactor of how captions work. Logic from Transcripter
was refactored into 2 different classes: CaptionFile, which is the model that
retrieves raw SRT from S3, and provides accessors to different formats of the
captions file. The CaptionConverter class does the actual transformations from
srt format to webvtt and html.

Then the TranscriptController responds to file extensions for .vtt, .srt, and
.html. For captions in the <video> element, we now simply refer to a URL such
as "/transcripts/cpb-aacip_123-4567.vtt" and it works.

This commit also introduces WebMock which is used to mock request/responses to
S3.

Closes #1054.


afred:
@Muraszko don't worry about reviewing all this, but please read the description above, and if that looks good, please merge. Thanks!

Issue #1063, opened 2016-09-29
Poster image does not appear on initial page load; only after a page refresh.


afred:
**Done when** poster images for videos appear on initial page load.

afred:
So.. different behaviors with different browsers... of course...

**Chrome:** Thumbs load on search results page, but not initially on the details page. After refreshing the details page, poster images show up.

Super-annoying-caveat is that this problem goes away when you have web developer tools open. So it's hard to see what response code is being returned for the failed request to S3 -- I still have yet to see what it is.

**Firefox:** images took a long time to load on the search results page. But they _did_ load when you click through to the details page.

**Safari:** does not seem to have a problem. Images load on search results page, and details page.

caseyedavis12:
@afred: I think this can be closed.

PR #1067, opened 2016-10-19
Adds video-js back in as the media player.

### LINKED ISSUE #1043: Impove seek capability in player for long audio and video
**problem** for long audio and video, the playback progress bar is still pretty short, which makes it hard to seek to a specific part of the media

**Done when** users can easily seek to a specific Minute/Second combination for media that is up to an hour long.

### LINKED ISSUE #1064: Caption button missing in Firefox
@afred Video records with caption files on demo.aapb.wgbh-mla.org don't have a CC button to toggle on the player in Firefox (tested with FF version 48.0.2).  [See example here.](http://demo.aapb.wgbh-mla.org/catalog/cpb-aacip_394-78gf257f)

**Done looks like**
Video player in Firefox has a CC button to toggle captions on or off.
cc: @caseyedavis12

afred:
We were using video-js before, but removed it in favor of using native HTML5,
for the sake of fewer dependencies. However, we lost poster imgs for Audio
playback, and we had inconsistencies with closed-caption button in FF. So we
decided to use video-js again.

Note that we're pulling video-js scripts and styles from their recommended CDN
rather than serving it from our application. I did this because we don't have
a CDN set up for serving assets, and including video-js scripts and styles in
our precompiled assests would bloat them quite a bit. I'm guessing (without
any benchmarks) that serving from their CDN would be faster. This does, however,
make us reliant on the VJS links working for as long as we need them. If it they
turn out to be unreliable, we can always pull a copy down and serve it
ourselves.

Closes #1064. Closes #1043.


afred:
@Muraszko @caseyedavis12 pushed this branch to demo to fix the player issues. Please check it out...

**video w/out captions** - http://demo.aapb.wgbh-mla.org/catalog/cpb-aacip_15-df6k06x38b
**video w/ captions** - http://demo.aapb.wgbh-mla.org/catalog/cpb-aacip_41-67jq2t1w
**audio** - http://demo.aapb.wgbh-mla.org/catalog/cpb-aacip_16-c53dz03d53

Please give me a 👍 if it looks good.

afred:
this may also fix #1063.

caseyedavis12:
Awesome!! It looks great and works really well. The only thing I noticed is that in Firefox for a captioned file, there are “caption settings” and the user can change the font style, color, background, etc. This option is not available for Chrome and Safari. This is not a blocker, though.


@muraszko @afred

PR #1068, opened 2016-10-20
Adding spanish language genre term

sroosa:
I added the new genre term "Spanish Language" to the vocab map and to the browse by topics list on the homepage. I also added any of the missing genre topics or formats to the vocab maps (terms that we've added since the last time the vocab map was updated).


afred:
@caseyedavis12 @Muraszko going to get this test passing, then will merge. Stand by.

afred:
gah.. the test failure doesn't seem to have anything to do with your changes @sroosa. But, I still want to get it passing.

afred:
Well damn. I guess it has more to do with your changes than I thought.

By adding "News" to `genre-type-map.yml`, it is now both a topic and a genre. Is this what you want?

The test is failing, because the sample PBCore assumes "News" is a topic, not a genre.

If "News" is supposed to be a valid **genre** _and_ **topic**, then the way it's currently written, it will _always_ be converted to `<pbcoreGenre annotation="genre">News</pbcoreGenre>` and never to `<pbcoreGenre annotation="topic">News</pbcoreGenre>`.

However, I don't think we want to end up with...
   ```
   <pbcoreGenre annotation="genre">News<pbcoreGenre>
   <pbcoreGenre annotation="topic">News<pbcoreGenre>
   ```
... either, do we?

afred:
Actually, i just saw in the code where there _is_ logic to accomodate a term that is both a topic and a genre. I'll assume that that's what we want, fix the test, push, and merge.

sroosa:
@afred I think that's what i want.

PR #1071, opened 2016-10-26
Add 2 new orgs

sroosa:
We added collections from 2 new organizations in the AMS, but to get onto the AAPB website, they needed to have md pages. The page for Vision Maker Media is fully fleshed out. The page for Ken Burns - Florentine Films is more bare bones because we don't have the complete information from the organization; however, it should work for now and can be easily updated in the future.


afred:
@sroosa @Muraszko do we want this merged and pushed to live? Today, or when?

sroosa:
@afred yes, and as soon as possible. We have something that's supposed to start on Nov. 1st that's contingent upon these new orgs showing up on site.

PR #1093, opened 2016-11-10
Adds new dir for exhibit 'First Amendment'

### LINKED ISSUE #1092: Add cmless directory for "first-amendment" exhibit
**done when** persistent directory at "app/views/exhibits/first-amendment/"

afred:
Closes #1092.


afred:
@caseyedavis12 confirm directory name by checking the "Files changed" tab above. If correct, and once tests are green, feel free to merge using the "Create a merge commit" option.

afred:
fyi, the `.gitkeep` file is a way to create effectively "empty" directories in git. One of git's shortcomings is that it does not track empty directories, so you have to put a dummy file in there, conventionally named `.gitkeep` to indicate that its only purpose is to tell git to track the parent directory, whether there are any additional files in it or not.

After you put real files in the directory, and expect them to stay there, you can remove the `.gitkeep` file.

PR #1112, opened 2016-12-01
1055 Highlight Captions

jasoncorum:
Hey @afred! New pull request.

The approach is generally the same that we discussed earlier this week, but I moved the model methods to caption_file.rb. Main benefit being that we won't need to worry about about updating the ValidatedPBCore model and I think it makes sense for the logic to exist on a captionfile associated with the pb_core object. Then I had to do a lot of cleanup in the spec files to handle the new fixture. The new fixture was needed because the only caption file in the existing pb_core xml fixtures was a couple lines long.


app/views/catalog/_index_header_default.html.erb
```diff
@@ -41,12 +41,26 @@
                 <dd><%= date %></dd>
               </dl>
             <% end %>
-              
+
             <dl class="clearfix">
               <dt>Organization</dt>
               <dd><%= pbcore.organization.short_name %></dd>
             </dl>
```
jasoncorum: I don't think so. This update came from some existing tests around empty query submissions. So, this should only display captions if there a query is supplied and if the pbcore document has captions.

app/views/catalog/_index_header_default.html.erb
```diff
@@ -59,4 +73,4 @@
 
   </div><!-- END Document Header -->
 
-<% end %>
+<% end %>
```
afred: i don't care. I never thought about it until Rubocop bitched at me. The justification is for cat'ing onto the end of files consistently. So... meh. Forget about it.

spec/models/caption_file_spec.rb
```diff
@@ -55,6 +55,48 @@
     end
   end
 
+  describe 'catalog index displays highlighted captions in results' do
```
jasoncorum: Cool. I'll make the update.

app/views/catalog/_index_header_default.html.erb
afred: But there could be a query supplied _and_ the returned result might have caption, but the search query could be matching on the something other than the captions... i.e. title or description or something. That means the captions would show up in the search result, but nothing would be highlighted there, correct? Or am I missing something?

jasoncorum: Yes, that would be the case. Thanks for catching that! I'll refactor.

afred:
@jasoncorum I don't think i used Github's review feature the way i intended, i'll have to read more on how to use it. When you push the latest commits, just @ me again.

jasoncorum:
@afred, ready for another look.

PR #1114, opened 2016-12-02
Adds .present? to conditional in _index.header_default.html.erb

jasoncorum:
@afred, in checking to see if we already addressed issue 1076 (audio player display issue) and pulling master, I realized that I needed a .present? on the params[:q] conditional in the view.


afred:
good catch!

PR #1115, opened 2016-12-09
1035: Cleans Out Downloads dir after Ingest

jasoncorum:
@afred: I did a pretty significant refactor of download.rb, so please give that a thorough review. Also updated test to look for 2 folders before and after ingest. One weird thing I encountered was a link (https://www.statearchivists.org) that failed during the test run in Travis, but not locally, and it looked like an SSL error on Travis. The site itself, if you view the source, shows that it is serving up mixed content. Tried to troubleshoot, but could not get a solid solution, so I added to existing linkchecker ignore file.

### DIFF
--- a/scripts/download_clean_ingest.rb
+++ b/scripts/download_clean_ingest.rb
@@ -28,9 +28,9 @@ class DownloadCleanIngest
   end
 
   def download(opts)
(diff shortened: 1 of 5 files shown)


afred:
looks really good!

PR #1138, opened 2017-01-04
Updates home featured gallery with new content

caseyedavis12:


afred:
@jasoncorum this is something we'll have to look into. The "bad" URLs Travis is whining about are `/` and `https://www.statearchivists.org`, both of which are valid.

It's also complaining about the home page missing a required field, but the error msg doesn't say what it is. This could be due to leading spaces for some of the entries.

The error messaging here could be better, i.e. don't dump out the entire XHTML doc if one thing is found to be invalid. And if CMLess is missing an expected key, then please tell us which one it is. Would not be opposed to making a couple of tickets to make it better.

jasoncorum:
@afred - removed the extra spaces in home-featured.yml and that seemed to do the trick.

PR #1173, opened 2017-02-15
hard coded blurbs

sroosa:
for states with multiple NAEB stations

### DIFF
--- a/config/states.yml
+++ b/config/states.yml
@@ -48,8 +48,20 @@
 
 - state: California
   blurb_fill:
-    naeb_station: KPPC
-  blurb_md: *naeb_blurb
+    naeb_station:
+  blurb_md: >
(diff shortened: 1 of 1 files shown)


sroosa:
I tried to hardcode blurbs into states.yml, but maybe I didn't alter it correctly. I don't have values in the naeb_stations: section. Should I just delete that since it's hardcoded into the blurb, rather than being filled in programmatically?

afred:
@sroosa looks like [a syntax error in `states.yml`](https://travis-ci.org/WGBH/AAPB2/builds/201872925#L341)

If you're doing multi-line in YAML, you need to use YAML's folding syntax, which uses a `>`. See the Kansas blurb for an example.

sroosa:
@afred I can't figure out exactly what I did wrong. Can you check it out again? Sorry!

afred:
@sroosa Sorry for the delay in getting back to this. It's [a different error this time](https://travis-ci.org/WGBH/AAPB2/builds/202623497#L341). Looking into it.

afred:
@sroosa found it. TL;DR, unfortunately, you can't put links in the blurb that have the `%` character in them.

The `%` character is part of an escape sequence for URLs to handle special characters.

In our case, the links you were providing within `blurb_md` are canned searches, which, when copied from your browser, will have special characters like brackets and parentheses escaped using the `%`.

Normally this wouldn't be a problem, the `blurb_md` goes through a pre-processing step, where it does a string replacement for anything it finds of the form `%{replace_me}`.

It's the `%` signs from the escaped characters in the URL that were confusing the pre-processing step, and causing the cryptic error.

PR #1208, opened 2017-03-15
1134: Online Reading Room Rules of Use

### LINKED ISSUE #1134: Make ORR Rules of Use a window instead of full page
From Google Analytics, one can tell that some users stop at the ORR Rules of Use and do not accept to view the media. This may be because the page is confusing to users who do not know what to do on this page. 

We think that more people would click through the ORR Rules of Use if it appears as an pop up window hovering over the record page (slightly dimmed in the background) for the item that the user clicked to view. The Rules of Use would appear with a scroll bar and Accept Rules of Use would appear on the outside of the Scroll Bar. 

**Done when**
- [ ] ORR Rules of Use is presented to users as a hover window with the ORR rules of use scrollable within the window instead of full page.

Here is an example:

popupwindow

jasoncorum:
@afred - This was approved in the ticket thread. I didn't pull out all the infrastructure around the original functionality. Is that something you'd like included in this part of the work?

Closes #1134


jasoncorum:
@afred - update made!

PR #1233, opened 2017-04-12
454: Facebook Metadata

jasoncorum:
@afred - ready for review.


afred:
@jasoncorum looks good. I added a commit to have `PBCore#img_dimensions` cache the result from `FastImage.size` so it only get's called once per request to `catalog#show`. I think it was getting called 4x (2x each for `#img_width` and `#img_height`).

PR #1243, opened 2017-04-14
Fixes state abbr in states.yml

### LINKED ISSUE #1242: Wrong abbreviation for Washington state
Causes `undefined method `name' for nil:NilClass` error right here: https://github.com/WGBH/AAPB2/blob/master/app/views/catalog/_organization_facet.html.erb#L4-L6

**Done when:** abbreviation is corrected, no more error.

afred:
Closes #1242.


afred:
@jasoncorum review and merge if/when green?

PR #1268, opened 2017-06-02
1265 srt exception

### LINKED ISSUE #1265: Parse SRT in CaptionConverter Causes 500 on Search Index
In #parse_srt in CaptionConverter we raise an exception if there's an error in parsing the CaptionFile. since we're using this in our catalog index view, we need something a little more graceful.

jasoncorum:
@afred. Closes #1265.

### DIFF
--- a/lib/caption_converter.rb
+++ b/lib/caption_converter.rb
@@ -44,7 +44,7 @@ class CaptionConverter
 
   def self.parse_srt(srt)
     SRT::File.parse(srt).tap do |parsed|
-      raise InvalidSRT.new(parsed.errors) unless parsed.errors.empty?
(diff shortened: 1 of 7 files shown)


afred:
@jasoncorum So this will just blank the HTML if there's a parsing error?

jasoncorum:
@afred - Yep. Sorry, should have put in more context. A couple of additional points of information. First, I think the only code actually using the caption to html method is the code to present part of the caption in catalog#index. That was one of the first things I coded, I would probably rewrite that code now that I now more about parsing SRT files. I'd like to make another ticket for that. Two, since we know that the CaptionFiles are suspect at best, I think returning nil until instead of raising an error makes more since we're using the captions more frequently now and the code for for caption#index can handle a nil response.

PR #1278, opened 2017-06-21
Newsmagazine exhibit

sroosa:
Alejandra has finished her exhibit, so this is ready to be merged.


afred:
@sroosa This is getting hung up on a link that the tests say is bad, but I'm able to access it: http://www.clir.org/pubs/reports/pub164/pub164.pdf

sroosa:
@afred Any idea of this is the tests problem? and how to fix that?

afred:
I can't reproduce this bug locally. The relevant line in the error output is this..
```
[FAIL Peer certificate cannot be authenticated with given CA certificates http://www.clir.org/pubs/reports/pub164/pub164.pdf]
```
The `cURL` on my local machine is apparently different than the `cURL` installed on Travis... or at least they have a different set of CA certs their using.

Looking into ways of ignoring the CA cert check.

PR #1282, opened 2017-06-26
Upgrades to ruby 2.2.4

afred:
The latest Ruby version is actually 2.4.1, but jumping that far will require
some dependency resolution, which I found by trying to go just to Ruby 2.3.x.
The reason we're going to 2.2.4 now is because we're changing how we install
ruby on production machines. Instead of installing from source, or using RVM,
we're simply installing the yum package 'ruby22', which brings in the latest
point release, in this case 2.2.4.

This is easier (and faster)from a provisioning standpoint, which is why we're
doing it. But it means that the yum package may move forward with a point
release, and get out of sync with the ruby version specified in the Gemfile.
The result would be that running 'bundle install' will fail, which is something
we should catch when deploying to demo instances (prior to deploying to
production). The fix is to bump the ruby versions in the Gemfile again, just
like this commit does.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,7 @@
 source 'https://rubygems.org'
 
 # Specify specific version of ruby with which the app is compatible.
-ruby '2.2.2'
+ruby '2.2.4'
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
 gem 'rails', '4.1.2'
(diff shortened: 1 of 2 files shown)


afred:
For some reason, Github is not updating this PR after I force-push an amended commit. Not sure why, because I'm pretty sure it used to work.

Anyway, closing this PR and re-creating a new one.

PR #1284, opened 2017-06-26
20170626 release fixes

jasoncorum:
@afred - @sroosa had me release the latest code to demo (54.211.129.30). In manual testing I found a couple things that need tweaks. First, somehow the sticky nav on the homepage was overflowing on the x axis. Second, the new player.js file wasn't available for transcripts after release. I think precompiling it from assets.rb initializer should fix the issue. 

Could you take a quick look and merge? Thanks!


afred:
👍

Issue #1264, opened 2017-05-30
Station logos on appearing not Participating Orgs pages


caseyedavis12:
Done when
- [ ] station logos appear on participating org pages

Example: http://americanarchive.org/participating-orgs/1723
http://americanarchive.org/participating-orgs/1724

jasoncorum:
Note for self. The logos are present on S3 with underscores in their filename. Somehow the logo_src in the app is returning the logo name without the underscore. Not sure what would have changed that.

jasoncorum:
@sroosa - Looks like this is a Markdown issue. Using underscores for the Logo filename between non-words works fine:

http://americanarchive.org/participating-orgs/1347

But what happens with recognizable words is that the Markdown renders the underscore as <em> tags and concatenates the logo filename which then doesn't exist on S3:

http://americanarchive.org/participating-orgs/1724

We can fix this on our end by escaping the underscore with a backslash "\". So, in the example with 1724, changing the Logo filename from:

rocky_mountain_logo.png

to:

rocky\_mountain\_logo.png

fixes the issue.

sroosa:
@jasoncorum that sounds like a good work around to me. Since we create new markdown files by copying existing ones and updating the data, I don't think remembering to do this will be a problem.

Can we programmatically change all of those, or do I need to go into every md file?

PR #1286, opened 2017-06-27
Station logos

sroosa:
escapes the underscores in the logo filenames, so they are called from s3 correctly as per #1264


jasoncorum:
Looks good! 👍

PR #1292, opened 2017-06-28
Plays video from timecode included in URL

### LINKED ISSUE #1266: Create link capability to point to specific timecode in media file
In AAPB curated exhibits or elsewhere, we would like to be able to use a hyperlink to point to a specific time code in a video or audio file. If someone clicks on the link, it will take the user to the record page to a specific time code in the media file.

We should check to see if we can link to a specific chapter heading using webvtt.

jasoncorum:
@afred - This autoplays video on catalog#show if timecode included in URL. It's an extension of the transcript play button functionality. Closes #1266

### DIFF
--- a/app/assets/javascripts/player.js
+++ b/app/assets/javascripts/player.js
@@ -85,10 +85,11 @@ $(function(){
     });
 
     var url_hash = location.hash.match(/#at_(\d+(\.\d+))_s/);
+
+    // If timecode included in URL, play from that start time.
     if (url_hash) {
         $player[0].currentTime = url_hash[1];
-        // Autoplay generally a bad idea, but we could do it...
-        // $player[0].play();
+        $player[0].play();
     }
 
     // New for AAPB.
@@ -100,7 +101,7 @@ $(function(){
     var search = $('div.transcript-search');
     var searchButton = $('#transcript-search-btn');
 
-     function updateTranscriptGrid() {
+    function updateTranscriptGrid() {
       if ($divTranscript.hasClass('col-md-2')) {
         $divTranscript.addClass('col-md-6').removeClass('col-md-2');
       } else if ($divTranscript.hasClass('col-md-6'))  {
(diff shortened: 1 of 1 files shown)


afred:
@jasoncorum Are we sure we want to autoplay, in addition to jumping to the specified timecode?

jasoncorum:
@afred - I checked with @sroosa on it. The thinking is that if we're linking to a specific spot from somewhere the autoplay is preferred to getting to the page, seeing the thumbnail which may not correspond to the the specific clip we're linking to, and then clicking play where it will play from that timecode.

sroosa:
@afred @jasoncorum Casey also thought autoplay would be better.

afred:
Does the poster image still display after this line is executed? $player[0].currentTime = url_hash[1];`

Also, am I alone here in my hatred for autoplay of media?

sroosa:
@afred no we all hate autoplay. It's just that the only time someone would click one of these links  (I presume, which might be my problem) is if their trying to go see something specific right away. Like: I'm reading this exhibit. It talks about this one part in this program. I click on this link to see what it's talking about.

afred:
Well, the creator of the link isn't likely to be the follower of the link. I think if a user follows a link and sees that the media player is paused somewhere in the middle, then they will get the idea pretty quickly.

Don't mean to be a stickler over something rather small.. but if we all hate autoplay, why do it?

How about we make a point to all give it a shot on demo, with and without the autoplay, and see what feels better? In the meantime, I can merge this as-is. Sound good?

jasoncorum:
When the user clicks the link they get the thumbnail. When they click play they get the video at the time code.

One thought, though. I'm wondering if we can auto play and then pause right after to get the user to that point in the video and paused (rather than the thumbnail). 

I can check that out and double-check that I'm not missing another option.

jasoncorum:
@afred - i did a little digging this morning and it seems like the best way for us to get the user flow that we want to is to play the video to get past the thumbnail, jump to the timecode we want, and then pause the video. ran by @sroosa and @caseyedavis12 this morning.

PR #1299, opened 2017-06-30
Newsmag exhibit final tweaks

sroosa:
should fix the main section from exhibit page appearing at the end of each sub-section page.


jasoncorum:
Closing without merging because we thought there was an issue, but there wasn't. Just required a cache clear.

PR #1303, opened 2017-07-06
1116 Chapter Files

jasoncorum:
@afred - This is ready for review. It checks for a chapter file on s3 and displays the chapter button in the VideoJS player if one exists. One thing I don't like here is that I kept getting false HTML validation errors on the new fixture, so I included an option to bypass it (line 447 in catalog_spec). Essentially it kept returning that the video element doesn't have a closing tag, but it does and I could see it in the HTML returned. If you have thoughts on a better implementation for this ticket, they would be much appreciated!

### DIFF
--- a/app/models/chapter_file.rb
+++ b/app/models/chapter_file.rb
@@ -1,0 +1,18 @@
+require 'open-uri'
+
+class ChapterFile
+  URL_BASE = 'https://s3.amazonaws.com/americanarchive.org/chapters'.freeze
+
+  def self.vtt_filename(id)
+    "#{id}.vtt".gsub('cpb-aacip_', 'cpb-aacip-')
+  end
+
+  def self.vtt_url(id)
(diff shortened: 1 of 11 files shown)


app/models/chapter_file.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def self.vtt_url(id)
+    "#{ChapterFile::URL_BASE}/#{id}/#{vtt_filename(id)}".gsub('cpb-aacip_', 'cpb-aacip-')
+  end
+
+  def self.file_present?(id)
+    return true if Net::HTTP.get_response(URI.parse(ChapterFile.vtt_url(id))).code == '200'
+    false
+  end
+end
```
jasoncorum: @afred - Correct on both! Just removed.

afred:
👍 from me. Rebased to tidy up commit log. Waiting for Travis.

PR #1374, opened 2017-10-25
1334 special collections

caseyedavis12:


jasoncorum:
@afred - This is ready for review. I have a few more commits in there than usual because this PR closes a number of tickets, so wanted to be able to track the commit back to the issue in Github. Let me know if you have any questions on this PR!

PR #1417, opened 2018-01-16
Naeb special collection

caseyedavis12:
@jasoncorum @sroosa please review this new special collection


jasoncorum:
@caseyedavis12 @sroosa - Checking out the PR now.

jasoncorum:
The failed checks look like false positives from the LinkChecker. I can fix that in review after we address the following. In reviewing what the changes look like, I saw a couple of things. 

1. The design wasn't built to handle the number of characters we're using in the Featured Items section. I believe this was something we flagged during design as something we'd have to manage. We'll need to substantially shorten the titles used here.

2. The ready-made audio thumbnail doesn't quite work since it has the "Audio" text on it. How would you like to handle? I could try to re-crop the audio thumbnail and we can see how that works? Or we could do something else entirely. Thoughts?

screen shot 2018-01-16 at 11 05 18 am

caseyedavis12:
@jasoncorum  - can we just use this instead?
aabp_color pixel sq_no a

The shortened titles look OK to me, except I think I had a typo in Tennessee Williams name. Can you add an extra e at the end of his name?

caseyedavis12:
Here is a png of that same image. For some reason the colors are off in the jpg.

aabp_color-pixel-sq_no-a

jasoncorum:
@caseyedavis12 @sroosa - here is what that thumbnail looks like. if we're good to go, i can squash the commits and merge the request. thanks!

screen shot 2018-02-16 at 10 08 13 am

sroosa:
@jasoncorum This looks good! Let's go with it!

PR #1435, opened 2018-02-16
Creates NAEB Special Collection

jasoncorum:
I'm ready to merge this PR @caseyedavis12. But one more question before I do.

I noticed that there was a "special thank you" listed in the optional Funders section that read: Special thanks to Stephanie Sapienza for her contributions to the curation of this collection.

The Funders section was designed to be a logo with some text, so this wasn't rendering at all on the Collection. I moved it to the bottom of the Collection Summary collection. Is that cool? If yes, then I'll get this thing merged. Thanks!

### DIFF
--- a/app/views/special_collections/naeb.md
+++ b/app/views/special_collections/naeb.md
@@ -1,0 +1,41 @@
+# National Association of Educational Broadcasters Programs
+
+## Thumbnail
+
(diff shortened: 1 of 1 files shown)


caseyedavis12:
@jasoncorum: can you instead put it at the bottom of Collection Background?


best,

Casey

PR #1456, opened 2018-03-22
Completes video loading background

### LINKED ISSUE #1310: Make record page loading look better
With the new layout, when a record page loads, the play button from the video player and the bar of the transcript display window appear and then it takes a long time for the video to load and the page to snap into it's intended layout. What can we do to make this look better (either faster loads or a graphic to display while loading or something else?)

jasoncorum:
* Closes #1310

### DIFF
--- a/app/assets/stylesheets/locals/catalog.css.scss
+++ b/app/assets/stylesheets/locals/catalog.css.scss
@@ -256,7 +256,6 @@ dl:after {
 }
 
 .player {
-	padding: 0 15px 0.5em 0;
 	text-align: center;
 }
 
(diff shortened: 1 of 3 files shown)


afred:
@jasoncorum it looks like you're putting the transcript and the player on different rows, where it wasn't before. Does this affect the display, and if so, did you run those visuals by @sroosa and/or @caseyedavis12?

afred:
Also, needs some conflict resolution

afred:
Gotcha @jasoncorum, the viewing the diff in "split" view instead of "unified" makes is a bit clearer. My brain's renderer isn't powerful enough to know what it looks like, but if you got  👍 on the visuals, then super! Merging.

PR #1477, opened 2018-04-25
1471 Multiple Organizations + States

jasoncorum:
@afred - This PR allows for multiple organizations and states to be associated with a record. As a result of this update, we are now using a new XML validation file, updating the view, and had to tweak how we're generating mods xml to conform with the new PBCore standards.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -62,21 +62,28 @@ class CatalogController < ApplicationController
     config.add_facet_field 'media_type'
     config.add_facet_field 'genres', label: 'Genre', solr_params: { 'facet.limit' => -1 },
                                      message: 'Cataloging in progress: These tags do not reflect all AAPB content.'
-    config.add_facet_field 'topics', label: 'Topic', solr_params: { 'facet.limit' => -1 },
-                                     message: 'Cataloging in progress: These tags do not reflect all AAPB content.'
+    config.add_facet_field 'topics',  label: 'Topic',
+                                      solr_params: { 'facet.limit' => -1 },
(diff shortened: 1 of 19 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -62,21 +62,28 @@ class CatalogController < ApplicationController
...
-                                   message: 'Cataloging in progress: only half of the records for digitized assets are currently dated.'
-    config.add_facet_field 'access_types', label: 'Access', partial: 'access_facet',
-                                           tag: 'access', ex: 'access', collapse: false
+    config.add_facet_field 'year',  sort: 'index',
+                                    range: true, message: 'Cataloging in progress: only half of the records for digitized assets are currently dated.'
+    config.add_facet_field 'access_types',  label: 'Access',
+                                            partial: 'access_facet',
+                                            tag: 'access',
+                                            ex: 'access',
+                                            collapse: false
```
afred: @jasoncorum confirm this is just rearranging?

jasoncorum: Yep. For some reason it started getting flagged in rubocop for spacing issues. Didn't change anything.

app/models/organization.rb
```diff
@@ -81,6 +81,10 @@ def self.find_by_state(state)
     @orgs_by_state[state]
   end
 
+  def self.organizations(organizations)
+    organizations.map { |org| Organization.find_by_pbcore_name(org) }
+  end
+
```
afred: @jasoncorum what is the parameter here?

jasoncorum: That could probably be better named as `organization_names`, which should just be an array of the organization names pulled from the xml.

jasoncorum:
@afred - should be good to go now.

PR #1479, opened 2018-04-30
Limits IDs in view to AAPB and NOLA

### LINKED ISSUE #1421: Only show AAPB ID and NOLA Codes
Rather than listing every local id in the record, the display on the left of the record page should only show the AAPB ID and a NOLA Code, if the source of the ID includes NOLA. Otherwise, the ID should not appear here. 

If we decide that we want some of the ids back, we will work on deciding which ones on case by case basis.

jasoncorum:
@afred - this PR limits the IDs shown in the catalog#show view to just AAPB or NOLA ID codes. In particular take a look at the change in catalog_spec.rb please. This was the easiest/fastest way I could think to get around the issue of using our to_solr method to test our views. But I definitely think that is something we should change.

* Closes #1421

### DIFF
--- a/app/models/pb_core.rb
+++ b/app/models/pb_core.rb
@@ -102,6 +102,9 @@ class PBCore # rubocop:disable Metrics/ClassLength
   def ci_ids
     @ci_ids ||= xpaths("/*/pbcoreIdentifier[@source='#{SONY_CI}']")
   end
+  def display_ids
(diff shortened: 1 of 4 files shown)


jasoncorum:
@afred - not ready for merging. need to make an update based on a conversation with @sroosa: 

[11:05 AM] Sadie Roosa: can you have it display the id if NOLA/nola is anywhere in the source attribute
[11:06 AM] Sadie Roosa: not just if it's NOLA Code specifically?

jasoncorum:
@afred - ready to roll!

PR #1520, opened 2018-07-03
Hides Featured Items if None Included

jasoncorum:
Also fixes funder for NET.


jasoncorum:
Note to future us: we're merging this regardless of the failed test because the broken link will be available in the future.

@sroosa - could you review and merge?

PR #1527, opened 2018-07-17
Update show.html.erb to hint at LRA

caseyedavis12:
Update show.html.erb to hint at LRA


caseyedavis12:
@sroosa Alan, having returned from his sabbatical, does not agree with this change on the website. Can you roll it back for now and include in the next redeploy?

PR #1559, opened 2018-08-24
close-files-after-ingest------dont-delete-fixtures

foglabs:


jasoncorum:
@foglabs - i did a quick fix on some rubocop errors, but could you take a look at the failing test on zipper_spec.rb?

PR #1561, opened 2018-08-24
1534 NAEB Links

sroosa:
Adds a new variable to the blurb in the state.yml and populates that variable for multiple states with updated values.


sroosa:
This isn't doing what I want it to. Closing it to try again.

PR #1562, opened 2018-08-27
1534 NAEB Link Fixes

sroosa:
updates the relative link in the naeb_blurb to include the naeb special collection param, and updates the search terms


sroosa:
I'm abandoning this approach. Will try again slightly differently later.

PR #1563, opened 2018-08-28
1557 move linkchecker out of suite

foglabs:


app/mailers/notifier.rb
```diff
@@ -0,0 +1,17 @@
+class Notifier < ActionMailer::Base
+  default from: 'jason_corum@wgbh.org'
```
jasoncorum: I feel like we should probably not use our emails in this and try to get a generic email distribution list for the MLA developers to receive and I think we have an AAPB generic address we can use for the from.

foglabs: good point!

spec/support/validation_helper.rb
```diff
@@ -53,7 +55,18 @@ def expect_fuzzy_xml(options = {})
...
+      filename = "link_checker_result_#{Time.now.strftime('%m.%d.%Y')}.csv"
+      CSV.open(filename, 'a') do |csv|
+        row = [current_path]
+        row += bad_urls
+        puts row
+        csv << row
+      end
+    end
+
+    true
```
afred: This is a little confusing to me. Are all the bad URLs going into a single row of a CSV?

PR #1564, opened 2018-08-28
1193 Upgrade to Ruby 2.5

jasoncorum:
Upgrades AAPB to Ruby 2.5 and Rails 4.2.8.


jasoncorum:
@afred - i was still getting permission errors when installing bundler via mla-playbooks despite getting past installing rvm correctly, so i'm looking in to using the amazon ruby 2.4 package now. the database timeout update was to address a sqlite3 database issue where it was locking on the saved searches from blacklight during testing. i think we were running in to a concurrent access issue in our tests.

jasoncorum:
Closing without merging in favor of: https://github.com/WGBH/AAPB2/pull/1566.

PR #1565, opened 2018-08-30
American experience 2

mivillesvik:
Its been a bit since the last time I issued a pull request, so let me know if I have to change anything!


sroosa:
@jasoncorum @foglabs Ready for review! Probably want to squash these commits. Seems to have gotten pretty hairy in there!

PR #1566, opened 2018-08-30
Upgrade to Ruby 2.4

jasoncorum:
Upgrades AAPB to Ruby 2.4 and Rails 4.2.8. Going with 2.4 since it is available as a package from AWS.


jasoncorum:
@afred - this is ready for review.

PR #1579, opened 2018-09-20
1577 s3fs

jasoncorum:
@afred - This adds the necessary components for storing the rails and ingest logs on s3 through s3fs.


jasoncorum:
@afred - i was thinking that would be additional tickets following this work because I feel like there's potential to move a lot of stuff over if this ends up working.

PR #1583, opened 2018-10-01
1496 add credits header

foglabs:


foglabs:
failing test here for a new pbcore method, fixing now

sroosa:
@foglabs I'm getting the Credits section looking like this.  
screen shot 2018-10-02 at 10 34 50 am

Can we update it to only include the list at the top, not the broken out sections that are still at the bottom?  And is it possible to have the list more indented?

caseyedavis12:
@sroosa: it doesn’t look like there is any order to the list below. Can it be ordered alphabetically or from most important to least important roles?

PR #1593, opened 2018-10-09
1339 caption/transcript snippets in search results 

foglabs:


app/models/transcript_file.rb
```diff
@@ -24,6 +25,23 @@ def html
     @html ||= build_html
   end
 
+  def snippet_from_query(query)
+    transcript = Nokogiri::HTML(html).text.delete("\n")
```
afred: Consider replacing `Nokogiri::HTML(html).text` with just `text` (i.e. there is a `#text` method on both `TranscriptFile` and `CaptionFile` that should get you here).

scripts/download_clean_ingest.rb
```diff
@@ -16,6 +17,8 @@ class ParamsError < StandardError
 end
 
 class DownloadCleanIngest
+  include ActionView
+
```
afred: Don't think we need this.

scripts/download_clean_ingest.rb
```diff
@@ -5,6 +5,7 @@
 require_relative 'lib/pb_core_ingester'
 require 'logger'
 require 'rake'
+require 'action_view'
```
afred: try moving this to wherever `ActionView` modules are included.

PR #1598, opened 2018-10-18
1467 link to fix it plus

foglabs:


app/models/solr_document.rb
```diff
@@ -17,4 +18,16 @@ class SolrDocument
...
+    # self[:xml].include?('Captions URL')
+    # Nokogiri::XML(self[:xml]).xpath('//pbcoreAnnotation[@annotationType="Captions URL"]').first
+    Nokogiri::XML(self[:xml]).css('pbcoreAnnotation[annotationType="Captions URL"]').first
+  end
+
+  def transcript?
+    # self[:xml].include?('Captions URL')
+    # Nokogiri::XML(self[:xml]).xpath('//pbcoreAnnotation[@annotationType="Captions URL"]').first
+    Nokogiri::XML(self[:xml]).css('pbcoreAnnotation[annotationType="Transcript URL"]').first
+  end
```
afred: don't need the comments in these methods

app/views/catalog/_mobile_transcript.html.erb
```diff
@@ -18,6 +18,12 @@
             <%= raw @transcript_html %>
           </div>
         </div>
+
+        <div id="transcript-message" class="">
+          <div class="mobile-transcript-content">
+            <%= @transcript_message %>
+          </div>
+        </div>
```
afred: wrap in a conditional `if @transcript_message`

PR #1602, opened 2018-10-26
1548 Add 10s skip buttons under player

foglabs:


caseyedavis12:
@foglabs, for accessibility purposes, can we have these buttons say "Forward 10 seconds" and "Rewind 10 seconds"


Here is what the LOC does: [cid:eccf90e4-a96a-44aa-9c49-8f2aaac18c62]

foglabs:
screen shot 2018-12-13 at 1 59 30 pm

foglabs:
screen shot 2018-12-13 at 3 21 53 pm

sroosa:
@caseyedavis12 the buttons use aria labels to tell screen readers that they are for Rewind 10 seconds and Forward 10 seconds. But the buttons themselves just use the commonly agreed upon iconography for these actions. I think this is the best of both worlds, but let me know if you want us to go back to using the full text on the visual buttons.

caseyedavis12:
@sroosa: I saw the update with the << 10s and 10s >> and it looks good. Thanks!

PR #1603, opened 2018-10-29
1183 fallback to vtt if no srt

foglabs:
Just needs a test for a vtt-sourcefile caption.


afred:
consulted with @foglabs who committed final PR to get tests passing

PR #1620, opened 2018-12-06
Remove logger dependency

foglabs:
-updated jettywrapper
-Replaced wordpress gem with WPData class to fetch wp posts


afred:
Ha... well damn. I approved it, but failed to see the failure first.

We may have to webmock out a response from the WP endpoint.

PR #1622, opened 2018-12-07
Update oeta-news-cultural-programming.md

mivillesvik:
OETA GUIDs were wrong (are now fixed)!


sroosa:
Consolidated into #1634

PR #1633, opened 2018-12-13
'no stub, yes assets, rails js tag'

foglabs:


config/environments/production.rb
```diff
@@ -47,11 +47,8 @@
   # Prepend all log lines with the following tags.
   # config.log_tags = [ :subdomain, :uuid ]
 
-  # Use a different logger for distributed setups.
-  # config.logger = ActiveSupport::TaggedLogging.new(SyslogLogger.new)
-
   # Only keep up to 10 log files of ~ 1MB each.
-  config.logger = ActiveSupport::Logger.new('aapb-remote-storage/log/production.log', 10, 1.megabytes)
+  # config.logger = ActiveSupport::Logger.new('aapb-remote-storage/log/production.log', 10, 1.megabytes)
```
afred: @foglabs is this related to this functionality? Is it solving something else? Unless this is breaking something specific, I think we want to keep it.

config/environments/production.rb
```diff
@@ -23,7 +23,7 @@
   config.serve_static_assets = false
 
   # Compress JavaScripts and CSS.
-  config.assets.js_compressor = :uglifier
+  # config.assets.js_compressor = :uglifier
```
afred: I think we want to keep the JS compressed in production, yeah?

PR #1638, opened 2018-12-17
Exhibits - Markdown Credits changes

foglabs:


foglabs:
I *believe* that the only thing failing at this point is catalog_spec#links to collection. The record referred to in that test does not appear to belong to a special collection (or at least has no special_collection annotation). So, may just be a matter of changing that test to refer to a different record (that is in a special collection).

PR #1643, opened 2019-01-03
content fixes for exhibits

sroosa:


app/views/exhibits/watergate/additional-resources.md
```diff
@@ -1,4 +1,4 @@
-# Resources
+# Additional Resources
```
jasoncorum: @foglabs - is this going to break the Exhibit attr_reader for resources?

foglabs: doh! yes it will

foglabs: psych! this is intended (per sadie)

PR #1651, opened 2019-01-07
'conditional backgroundposition for ex gal'

foglabs:


app/assets/javascripts/player.js
```diff
@@ -39,24 +39,6 @@ $(document).on('turbolinks:load', function() {
       return $player.data('user-scroll');
   }
 
-  function skipPlayer(forward) {
```
jasoncorum: @foglabs - just want to make sure this delete is intended since it looks like the new-ish 10s skip functionality.

foglabs: yep, this correct - there were two copies of the skipPlayer function, possibly from a rebase

PR #1757, opened 2019-06-25
Fix Transcript Sync Button Issue

foglabs:
Last chunk of a transcript section was being added on the wrong side of the `.play-from-here` timecode in transcript viewer.


lib/transcript_viewer_helper.rb
```diff
@@ -9,16 +9,19 @@ def build_transcript(transcript_parts, source_type)
...
+        transcript_parts.each_with_index do |part, i|
           new_end_time, text = timecode_parts(part, source_type)
           if (new_end_time - last_end_time) > 60
-            @buffer += text
             build_transcript_row(doc_root, last_end_time, new_end_time)
             last_end_time = new_end_time
             @buffer = ''
+
+            # text for this step is actually first chunk of next paragraph
+            @buffer += text
```
afred:
```
@buffer = ''
@buffer += text
# can be simply
@buffer = text
```

afred: Also, can we change `@buffer` from an instance variable to a parameter passed to `build_transcript_row`?

PR #1789, opened 2019-07-12
Completes Ability specs migrated from user_spec.rb

jasoncorum:


spec/models/abilities/pbcore/can_api_access_transcript.rb
```diff
@@ -0,0 +1,36 @@
...
+      it 'skip_tos returns true for a public record' do
+        expect(ability).to be_able_to(:access_transcript, public_pbcore_record)
+      end
+
+      it 'skip_tos returns false for a protected record' do
+        expect(ability).to be_able_to(:access_transcript, protected_pbcore_record)
+      end
+    end
+  end
+end
```
afred: Spec descriptions mention `skip_tos` but that method isn't being used directly in the spec. Can we switch the descs to something like "[allows|does not allow] access to [public|protected] record"... or something like that?

PR #1808, opened 2019-07-23
1769 Adds additional title cleaner rules

jasoncorum:
1. words with apostrophes in between letters get titlecased: WOMEN'S should be Women's
2. words with hyphens in between letters get titlecased: ANTI-COMMUNISM should be Anti-communism
3. words with periods in the middle should stay all caps: U.S. should stay U.S. not U.s.


sroosa:
if it's cool, I'm going to piggy back on this PR to also add a few more terms to the ALL CAPS list.

PR #1819, opened 2019-08-20
Moves title on SpecialCollection index

jasoncorum:
Moves title from an overlay to beneath the thumbnail. It also sets dimensions on the title so that the columns stack correctly.


jasoncorum:
@foglabs - yup! @sroosa approved the screenshot with no bg-color for the titles.

PR #1832, opened 2019-09-03
Adds Sorting by Episode Number

### LINKED ISSUE #1653: Sort by Episode Number
If a Found Set includes ```<pbcoreTitle titleType="Episode Number">``` allow sorting by Episode Number

jasoncorum:
Closes #1653

### DIFF
--- a/app/models/pb_core_presenter.rb
+++ b/app/models/pb_core_presenter.rb
@@ -101,7 +101,7 @@ class PBCorePresenter
     @title ||= build_display_title
   end
   def episode_number_sort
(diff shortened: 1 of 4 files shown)


app/models/pb_core_presenter.rb
```diff
@@ -100,6 +100,9 @@ def titles
   def title
     @title ||= build_display_title
   end
+  def episode_number_sort
+    @episode_number_sort ||= titles.select { |title| title[0] == "Episode Number" }.map(&:last).sort.first
```
foglabs: note to selves: the `PBCorePresenter#titles` method now gives out a hash instead of a tuple array in https://github.com/WGBH-MLA/AAPB2/pull/1834

PR #1834, opened 2019-09-05
1774 factories not fixtures

foglabs:


afred:
Any further work on this will take a different shape. Can always re-open if needed.

PR #1845, opened 2019-09-17
WAPINA special collection

sroosa:


sroosa:
Rebasing worked. Both of these tests are green in Travis, it's just not making its way over to GitHub. Merging anyway.

PR #1857, opened 2019-10-15
Update rock&roll.md

mivillesvik:
I'm not sure if the "&" in the URL is going to work, so if not I can change it to "rock-and-roll.md"!


caseyedavis12:
@mivillesvik: I think we'll need to change it to rock-and-roll and avoid the special character. Thanks for checking!

mivillesvik:
Ok, I can change the name and do another PR!

sroosa:
Changes merged in #1864

Issue #1855, opened 2019-10-11
Browse section URLs with quoted terms are broken


sroosa:
The Browse section (https://americanarchive.org/#browse) has a URL for each term that does a keyword search for a specified set of related terms. Some of these keywords are phrases in quotes, such as "stock market" for the Economics URL. At some point (my guess is when we added the ability to run quoted searches against an unstemmed solr field), the URLs that have quotes around phrases stopped bringing up any search results. Example: https://americanarchive.org/catalog?q=economics+OR+economy+OR+%22stock+market%22+OR+%22tax+reform%22+OR+business&f[access_types][]=digitized

We need to update these URLs so that they lead to actual search results again.

sroosa:
Closed by #1865

PR #1865, opened 2019-10-30
Fix Quote Handling To Pass Through 'Title' Advanced Searches

foglabs:


foglabs:
Basic search
-No quotes returns all matches from any field
-Quotes returns all matches of quoted clause from any field

Advanced Search
-Title search field returns only records whose title contains the exact quoted clause that was searched for

sroosa:
@foglabs does this code also fix the issue in #1855?

jasoncorum:
@foglabs - i approved but then thought: does it make sense to add a test for this?

PR #1892, opened 2019-11-12
Timecode Embed Firefox Fix

foglabs:


foglabs:
Summary of changes:

-Default to `#at_0.0_s` when player uninitialized (my bad)
-Try to set player time immediately if `$player` contains videojs player (for chrome)
-Hook into vjs `loadstart` event, search for player element, and then try to set time (for firefox)
-Restructure player.js code to ensure functions/handlers are defined for both of these cases

PR #1895, opened 2019-11-19
Adds CanonicalUrls

jasoncorum:
We have a small number of records where we'd like to add a canonical url to the head of the record's view. We decided to record these records in a YAML file and add to the view if there's a canonical url present.

### DIFF
--- a/app/models/canonical_url.rb
+++ b/app/models/canonical_url.rb
@@ -1,0 +1,33 @@
+require 'yaml'
+require_relative '../../lib/aapb'
+require 'active_support'
+require 'active_support/core_ext'
+
+class CanonicalUrl
+  # Only a small number of records will have a canonical_url, so they are stored in a YAML file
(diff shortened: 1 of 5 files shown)


app/models/canonical_url.rb
```diff
@@ -0,0 +1,33 @@
...
+  def find_url(id)
+    return nil unless ids_with_urls.include?(id)
+    canonical_urls.select { |url| url["id"] == id }.first["url"]
+  end
+
+  private
+
+  # Get all the canonical URLs from config
+  def canonical_urls
+    YAML.load_file(Rails.root + 'config/canonical_urls/url_map.yml')
```
foglabs: Everything looks good, but could we cache the file?

```
Rails.cache.fetch('canonical_urls') do
  YAML.load_file(Rails.root + 'config/canonical_urls/url_map.yml')
end
```

PR #1902, opened 2019-11-26
Stub Rails.cache method correctly for each method that uses it 

foglabs:
This is necessary because new code was causing the ingest to crash when it tried to access `Rails.cache`

Defines the guts of the cache method in `rails_stub`

*Wow!*


foglabs:
tested ingest script on demo, should be all set

PR #1934, opened 2020-01-22
Backstory

sroosa:


sroosa:
@mivillesvik the only issue with your Sort entry was that you needed a +asc or +desc for ascending order to descending order. i fixed it.

PR #2005, opened 2020-04-29
'adds logic to confirm transcript url before attempting to download'

foglabs:
Checks for .json transcript file, then .txt file if not found. Adds HTTParty gem, because curb/curl gem's `head` method seemed to attempt to download the file instead of sending a HEAD request.


jasoncorum:
Approved, but the cops aren't happy. You gonna have to talk them down.

PR #2040, opened 2020-05-21
Update black-journal.md

mivillesvik:


afred:
@mivillesvik @foglabs This is no longer needed, right?

caseyedavis12:
@afred: yes, we are launching both a "Black Journal" Special Collection and a Black Power exhibit. The Black Power exhibit will have a Black Journal section and a SOUL! section.

PR #2129, opened 2020-08-31
Update peoplenearhere.md

mivillesvik:
new special collection!


singlesoliloquy:
@mivillesvik this needs ## Funders as well.

PR #2136, opened 2020-09-21
Updates Iowa Public Television to Iowa PBS

jasoncorum:
Updates name for Iowa Public Television to Iowa PBS and adds some additional info. All the records with Iowa Public Television as an instantiationAnnotation with an annotationType of "organization" will have to have to be reingested with the update for the View All link to work.

### DIFF
--- a/app/views/organizations/md/1770.md
+++ b/app/views/organizations/md/1770.md
@@ -1,8 +1,8 @@
-# Iowa PBS
+# Iowa Public Television
 
 ## Short name
 
-Iowa PBS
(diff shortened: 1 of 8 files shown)


foglabs:
does the organization name change need to happen to IPT records in AMS2 first?

singlesoliloquy:
Yes. Need to re-ingest public television stuff.

afred:
### BLOCKED
Need to plan for full re-index.

ekemeyer:
Fixed with #2910

PR #2140, opened 2020-09-29
Fix &amp; displaying in citations 

foglabs:
Wrong method was used for printing out short name for contributing organizations, needed to explicitly gsub `&amp;` into `&` (as in `Organization#short_name`)


afred (approved):
👍

PR #2154, opened 2020-10-21
Add NewsHour appearances

caseyedavis12:
Thanks to the great work of intern Declan McBride. Let's launch this to demo today and to production tomorrow after QC in time for tomorrow night's debate. Let's also launch the Prospects of Mankind and People Near Here special collections which should be ready to go.


singlesoliloquy:
Ready for demo.

PR #2175, opened 2020-11-04
Focus 580

rynmarchese:


app/views/special_collections/focus-580.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Collection Background
```
foglabs: Whoops, the PR is failing because this heading needs to be just '## Background'

foglabs: I'm making the change, should be good after that.

PR #2179, opened 2020-11-10
2171 Rewrite/Reindex PBCore

jasoncorum:
ThePBCoreCorrector rewrites titleTypes and descriptionTypes that have been mistakenly indexed as a source attribute. The class takes an array of GUIDs and run the process.


foglabs (approved):
入って、ください

PR #2229, opened 2021-04-08
Metadata transcripts statement

caseyedavis12:
Adds new metadata and transcripts statement page, section of FAQ, link in Footer, and link in box that appears on record pages.


afred (approved):
👍

PR #2230, opened 2021-04-12
Moves Feedback Form to Top-level Navigation

jasoncorum:
We want more Feedback from users, so we're moving the link to the Feedback for to a top-level navigation item/


jasoncorum:
Screen Shot 2021-04-30 at 7 57 29 AM
Screen Shot 2021-04-30 at 7 58 30 AM

afred:
👍 forgot to formally approve, but it's all good.

PR #2235, opened 2021-04-30
Decreases font size for gallery header

afred:
For special collections on mobile, the titles are getting chopped off.
The real fix is to be able to nicely display full title regardless of length,
however there are some practical limitation to that. Shrinking the font size
works for many of the titles as a quick fix.


afred:
self-approving to rush a Friday afternoon deployment, because 🎸

PR #2240, opened 2021-04-30
Fixes image display in Whats New page

afred:
Wraps images in a span that acts like a div, to contain the image and text together. Need
to use span because that's what markdown allows (only inline elements). Then change it's
display to inline-block. Hehe. What a hack. Anything better is welcome.


afred:
Here's what it looks like with the changes.
image

afred:
self approving

PR #2243, opened 2021-05-04
Jeanwalkinshaw

mivillesvik:
Merging Jean Walkinshaw special collection with master! Also updated image URLs


caseyedavis12:
@mivillesvik does it make sense to sort this collection by date instead of title? And I see a few programs are highlighted in the collection summary. Could we hyperlink to those in the text and make that a strategy going forward as well? Thank you!

PR #2244, opened 2021-05-05
Links back to AAPB from Embed Open in New Window

jasoncorum:
When users embedded our player on other sites and our links didn't have the right target attribute, browsers would block content and display a security issue. This adds the attribute so that they open in a new window and don't cause the security issue.


afred (approved):
👍

PR #2251, opened 2021-05-11
Update the-murder-of-emmett-till-interviews.md

rynmarchese:


caseyedavis12:
@rynmarchese do we need to call out Teaching Tips? I didn't see we did that on the other resources and did for this one, but the resource includes a video and background essay as well as the Teaching Tips.

PR #2260, opened 2021-05-25
'static robots txt file'

foglabs:


afred:
pulled rank and merged in order to deploy asap. yay

Issue #2219, opened 2021-02-26
Adding more pins to New Mexico on AAPB map!


mivillesvik:
## User Story
Hi, this is pretty similar to #2215, but basically we're adding more stations to New Mexico for the New Mexico project so if possible can one of the devs please add more dots to New Mexico on the AAPB map? https://americanarchive.org/#stations

## Additional Information

New Mexico currently has one dot, but we'd love to have more to reflect that we have multiple stations from there!

![Screen Shot 2021-02-26 at 5.14.03 PM.png](https://images.zenhubusercontent.com/5fd38131b736e215633119d4/9a0163ae-6126-4def-96f2-be7be89b7b8f)

## Done Looks Like

Done looks like two more dots being on the AAPB organization map for New Mexico's state!

singlesoliloquy:
Big file and we put in coordinates

PR #2264, opened 2021-06-14
Adds pins for NM and KS

jasoncorum:
Adds pins for cities in KS and NM to the participating organizations map here: https://americanarchive.org/participating-orgs


foglabs (approved):

jasoncorum:
Closes #2219

Issue #2297, opened 2021-08-10
Use URI::escape instead of CGI::escape for generating title links in AAPN catalog#show


jasoncorum:
CGI::escape for PBCore titles in catalog#show is converting spaces in the titles to "+" which is causing issues when our newsletter system converts those "+" to "%2B" as that gets literally interpreted by SOLR as part of the title. URI:escape converts the space in a title to "%20" which works.

**Done When**
PBCore titles with spaces get escaped to %20 instead of %2B

PR #2298, opened 2021-08-10
Uses URI::escape instead of CGI::escape for links to titles

jasoncorum:
CGI::escape for PBCore titles in catalog#show is converting spaces in the titles to "+" which is causing issues when our newsletter system converts those "+" to "%2B" as that gets literally interpreted by SOLR as part of the title. URI:escape converts the space in a title to "%20" which works.

Tests pass locally except for these 2 which never do. 

Screen Shot 2021-08-10 at 1 38 40 PM


afred (approved):
👍

jasoncorum:
Closes #2297

PR #2315, opened 2021-08-30
Exhibit Link Index Fix

jasoncorum:
Updates the regex for how we index Exhibit subpages to a specific record. We had an issue where https links were not getting indexed. Also, in running tests, I found that the Conservatism exhibit notes page was broken by the introduction of the config. The new config would load but break when it tried to generate the thumbnail_url for the OG meta tags on the page. Removing extraneous sections from the notes.md fixed the issue.

These two failed 
Screen Shot 2021-08-30 at 11 31 59 AM
tests are my local ones that never pass:

### DIFF
--- a/app/models/exhibit.rb
+++ b/app/models/exhibit.rb
@@ -59,7 +59,7 @@ class Exhibit < Cmless
 
         # take custom config for this exhibit
         config = global_config[top_path].clone
-        
+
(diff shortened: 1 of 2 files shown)


foglabs (approved):

jasoncorum:
Closes #2316

PR #2357, opened 2021-11-02
Use HLS streaming for video playback

### LINKED ISSUE #2346: Use Ci Streaming Instead of Ci Download For AAPB Media
Because of the idea that using Sony Ci's streaming capabilities will make AAPB media less prone to being downloaded, we want to switch over.

Done when AAPB utilizes the `/streams` resource instead of `/download`

### LINKED ISSUE #2356: Resubmit streaming PR, excluding audio
### Because
We got streaming working for video in #2354, but had to revert it because it did not work for Audio.

### Background
I cannot find anywhere in the Sony Ci API docs where you can make an audio playback stream. I filed a support ticket with them, asking about it specifically. Until then, let's not use streaming for audio.

### Done when
1. Video uses streaming URL
2. Audio continues to use download URL.

afred:
Streaming from Sony Ci makes it a bit harder to capture media from our site, so we are
switching to using HLS streaming URLs rather than progressive download URLs.

To do this, we're switching to use the new Sony Ci API gem which provides a cleaner interface to the
underlying Sony Ci API, including an method for fetching HLS streaming URLs, and will allow for more
interaction with Sony Ci in the future.

NOTE: Sony Ci does not provide a streaming URL for audio files, so for now those remain as downloadable
URLs.

Also,
* removes a one-off script and spec dependent on the old Sony Ci API gem, but we don't need it anymore.
* guards against empty transcript file content when creating transcipt snippets in catalog#index
* guards against empty JSON param passed to the transcript view helper method

Closes #2356.


afred:
Self-review: 👍 . Justification: no reviewer availability until next week + time sensitive request.

PR #2409, opened 2022-04-06
Fixes error when transcript file is missing

### LINKED ISSUE #2408: BUG: ingest failing when constructed transcript URL not present
**Describe the bug**
in `PBCorePresenter#constructed_transcript_src`, if neither of the constructed Transcript URLs can be verified with a HEAD request, then method returns the array `%w(json txt)`, which in turn get's passed along as the URL for the `TranscriptFile` instance, and subsequently throws a `Bad URI` error when the `TranscriptFile` attempts to _also_ do a head request.

The `PBCorePresenter` can probably use `TranscriptFile` for determining existence of a real Transcript file at the URL.

So either we can do one of the following:
1. refactor `PBCorePresenter` to use `TranscriptFile` to determine existence of actual transcript file at the URL
1. tweak `PBCorePresenter#constructed_transcript_src` to return either the validated URL or nil (easier, but less DRY).

**To Reproduce**
Attempt to ingest a PBCore XML record using the `scripts/download_clean_ingest.rb` script, where the record does not have an existing Transcript file at the assumed location, e.g. `https://s3.amazonaws.com/americanarchive.org/transcripts/{GUID}/{GUID}-transcript.{EXT}`, where GUID is the record ID, and EXT is _either_ "json" or "text".

**Expected behavior**
Records do not fail as a result of not having a present Transcript file at the assumed transcript URL (`PBCorePresenter#constructed_transcript_src`)

afred:
closes #2408

### DIFF
--- a/app/models/pb_core_presenter.rb
+++ b/app/models/pb_core_presenter.rb
@@ -188,11 +188,13 @@ class PBCorePresenter
       # transcript guids are expected to have -
       trans_id = normalize_guid(id)
 
-      # check if s3 transcript is a json or txt
-      %w(json txt).each do |ext|
-        url = %(https://s3.amazonaws.com/americanarchive.org/transcripts/#{trans_id}/#{trans_id}-transcript.#{ext})
-        return url if verify_transcript_src(url)
+      # Map 'json' and 'txt' file extensions to full transcript URLs to look for.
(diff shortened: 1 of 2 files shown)


mrharpo (approved):
Looks good!

PR #2480, opened 2022-09-28
'use japan instead of india for geoip test'

foglabs:


foglabs:
doesnt work, still comes up as US

PR #2484, opened 2022-11-03
Change all "Watch from here" to "Access from here"

### LINKED ISSUE #2452: Change "Watch from Here" to "Access from Here"
## User Story
As a user, when I do a search of the AAPB and the keywords appear in a transcript, I should see a link that says "Access from Here" rather than "Watch from Here" since not all content in the AAPB is TV. 

Snap9 (1)

mrharpo:
# "Access from here"
Change `"Watch..."` to `"Access..."` because not all content is video. 

Also, the controller can't distinguish between the two without creating a new `PBCorePresenter` for each search result.

Closes #2452

### DIFF
--- a/app/helpers/snippet_helper.rb
+++ b/app/helpers/snippet_helper.rb
@@ -4,7 +4,7 @@ module SnippetHelper
     if timecode_url
       timecode_link = %(
         <a href="#{timecode_url}">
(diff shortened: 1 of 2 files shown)


afred (approved):
👍

PR #2486, opened 2022-11-04
Record title search link: show all results

mrharpo:
# Title search
On a record, when clicking on a search link from a title field, the link now includes `&f[access_types][]=all` to show all records, not just the search default of `online`.

Closes MLAAAPB-60


afred (approved):
👍

PR #2487, opened 2022-11-07
Add KERA (Dallas, TX) to `participating-orgs` map

### LINKED ISSUE #2345: Adding a dot for KERA to the AAPB organization map
## User Story
We've been working to add a lot more stuff from KERA in Texas and they've asked that they get a dot on the AAPB map (especially since they're in Dallas (north Texas) and the stations on the map right now are both in south Texas)!

## Additional Information
If it helps, here is their organization page on the AAPB: https://americanarchive.org/participating-orgs/NCSG55127

## Done Looks Like
There would be another dot on the state of Texas on the AAPB that represents KERA (Dallas).

mrharpo:
# KERA
Adds station dot for KERA in Dallas, TX

Closes #2345


afred (approved):
👍

PR #2496, opened 2022-11-29
`./aapb test` and `format` commands

mrharpo:
# `./aapb`
- Better docker command substrings for reusability
- Print `$HELP` if command is not recognized

## `./aapb test [FILENAME]`
- Runs rspec tests on file
- With no args, runs full test suite

## `./aapb format [FILENAME]`
- Runs `rubocop --autocomplete` on files
- With no args, runs on all files


afred (approved):
👍

PR #2497, opened 2022-12-01
Refactor geo-location service

afred:
* Removes sporadically failing tests relying on possibly outdated info.
* Renames GeoIPCountry to GeoLocation because we're not using geo_ip gem anymore.


foglabs (approved):

mrharpo (approved):
💪 Much better! Should have way fewer false negatives! ✅

PR #2501, opened 2022-12-02
Use prettier 404 page for missing catalog pages

afred:


foglabs:
I think this probably just needs the expected exception in the test changed as well

config/environments/test.rb
```diff
@@ -21,7 +21,7 @@
   config.action_controller.perform_caching = false
 
   # Raise exceptions instead of rendering exception templates.
-  config.action_dispatch.show_exceptions = false
+  config.action_dispatch.show_exceptions = true
```
mrharpo, line 24: Fixed tests by raising exceptions here. From https://stackoverflow.com/a/50020700/19192178

mrharpo (approved):
👍 Much prettier!

PR #2519, opened 2022-12-19
Exclude Notes Pages From 'Records in Exhibit'

foglabs:
Excludes notes.md pages from exhibit collation of 'these records are in this exhibit', and fixes broken guid links in several exhibits.


mrharpo (approved):
👍 Works on demo!

PR #2527, opened 2023-01-10
'revert ci video to non-streaming method'

foglabs:


mrharpo (approved):
🤞

afred (approved):
👍 emergency only! Still need to debug and revert this to use steaming url

PR #2546, opened 2023-02-10
Ignores more logs and prevents log crawling by bots

afred:
* Adds glob patterns to ignore Rails logs that have been rolled.
* Add /logs to robots.txt to prevent search engine indexing.


afred:
shameless self merge again, clean up a little bit missed from prev pr

PR #2606, opened 2023-06-02
Update 1770.md

mivillesvik:
updating short name per Marci Behm of Iowa PBS' request


ekemeyer:
Duplicate of #2893 and #2910

PR #2696, opened 2023-12-11
Handle json-breaking double quotes in pbcore values in JSON API

foglabs:
'cache pbcore xsl for json api, re-escape double quotes in xml before json transformation'


afred (approved):
👍

PR #2740, opened 2024-03-27
Adds new authorized referrers

### LINKED ISSUE #2739: Add U of Texas domains as authorized referrers
### Because
Request from U of Texas folks, et al, to use `/media/{aapb_id}` URLs in AVAnnotate tool.

### Background
Our User model defines a small list of authorized referrers that can access `/media/{aapb_id}` URLs directly.

NOTE: further configuration may be necessary for AVAnnotate tool to work as expected

### Done when
Domans for AVAnnotate tool are added to list of authorized referrers

afred:
Per request of folks from U of Texas using AVAnnotate tool, they want to be able to use /media/ URLs in their tools.

Closes #2739.


mrharpo (approved):
👍

PR #2749, opened 2024-04-16
🌌 Clusterize

mrharpo:
# Clusterize
Updates for Kubernetes deployment

- Sets AAPB_LOG env for production

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -51,7 +51,7 @@ Rails.application.configure do
   # config.log_tags = [ :subdomain, :uuid ]
 
   # Only keep up to 10 log files of ~ 1MB each.
-  config.logger = ActiveSupport::Logger.new('/var/www/aapb/current/log/production.log', 10, 1.megabytes)
+  config.logger = ActiveSupport::Logger.new(ENV['AAPB_LOG'] || '/var/www/aapb/current/log/production.log', 10, 1.megabytes)
 
   # Use a different cache store in production.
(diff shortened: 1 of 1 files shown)


lib/solr.rb
```diff
@@ -10,7 +10,8 @@ class Solr
   def initialize
     environment = ENV['RAILS_ENV'] || DEFAULT
     conf = YAML.load_file(Rails.root + 'config/blacklight.yml')
-    @connect = RSolr.connect(url: conf[environment]['url'])
+    solr_url = ENV['SOLR_URL'] || conf[environment]['url']
+    @connect = RSolr.connect(url: solr_url + 'blacklight-core')
```
afred, line 14: (see also above comment)

Blacklight gem may be expecting the core name to be a part of the configured URL, so we may run into issues by removing it from `blacklight.yml` and adding it back here.

scripts/lib/downloader.rb
```diff
@@ -112,7 +112,7 @@ def process_query(query)
     q = QueryMaker.translate(query)
     $LOG.info("Query solr for #{query}")
 
-    @options[:ids] = RSolr.connect(url: 'http://localhost:8983/solr/')
+    @options[:ids] = RSolr.connect(url: config.solr_url)
```
afred, line 115: (see other related comments)

we're leaving the core name off here, but adding it when connecting to Solr in `lib/solr.rb`.

afred (changes requested):

PR #2762, opened 2024-05-02
Adds script to parse apache logs looking for bad bots

afred:
# Usage
```ruby
ruby scripts/lib/apache_log_parser.rb glob/path/to/access_logs
```

## Caveats
* Running this script will not chante Apache config or restart Apache; you must do that deliberately.
* Running this script will only output the worst offenders in terms of number of requests within the log.
* The apache config it outputs can be cut and pasted by a privileged user into the apache config. After apache is restarted, the new config will take effect.
* This will only work using Apache's 'combined' log format. Any changes to the format will probably either break the script outright, or cause it to output invalid data and invalid Apache config. If apache log format changes, assume this script is broke.
* The ApacheLogParser object may be used within a Rails console to further analyze apache logs in interesting ways, i.e. by looking at agents, timestamps, requests urls, etc.

### Also
* Sets production log levels to 'warn' for Rails log.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -45,7 +45,7 @@ Rails.application.configure do
   # config.force_ssl = true
 
   # Set to :debug to see everything in the log.
-  config.log_level = :info
+  config.log_level = :warn
 
   # Prepend all log lines with the following tags.
   # config.log_tags = [ :subdomain, :uuid ]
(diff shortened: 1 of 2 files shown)


mrharpo (approved):
👍

foglabs (approved):

PR #2771, opened 2024-05-10
Update 1737.md

mivillesvik:
Need to confirm the following before attempting to repush records after name change takes place:

- [x] all records are in AMS2
- [ ] all records are valid and can be pushed from AMS2
47 of 16700 not valid - will manage these after pushing the valid records.


foglabs:
@mivillesvik this is annoying, but because this changes the shortname for South Florida PBS, this can't be merged until all records with `<instantiationAnnotation type="organization">WPBT2</instantiationAnnotation> have been updated to use the new name `South Florida PBS`

PR #2824, opened 2024-08-14
📟 API endpoint search for old-style IDs

### LINKED ISSUE #2773: BUG: JSON API endpoint for catalog records doesn't work with old-style AAPB IDs having underscore
### Replicate
* New style with dash works: https://americanarchive.org/api/cpb-aacip-512-707wm14j8q.json
* Old style with underscore doesn't: https://americanarchive.org/api/cpb-aacip_512-707wm14j8q.json

### Done when
Links with old-style IDs work.

### Background
The `IdHelper` class in `/app/helpers/id_helper.rb`  is used to solve this problem for non-API URLs:
* New style works: https://americanarchive.org/catalog/cpb-aacip-512-707wm14j8q
* Old style also works: https://americanarchive.org/catalog/cpb-aacip_512-707wm14j8q

Notice that we don't do redirecting here. There are pros and cons to not redirecting, but I think right now staying consistent within the code probably has more net value than enforcing the canonical new-style URLs through redirection. So let's stay consistent by not redirecting unless somebody has a more compelling argument for switching to a redirection strategy.

### LINKED ISSUE #2827: Use solr '?' wildcard to resolve /catalog/:id and  /api/:id routes
### Because

Because we currently resolve different guid styles by sequentially checking `-`, `_`, and then `/` guid variations on `/catalog/:id` until we find a match in solr, we can save ourselves from making multiple solr searches per request by subbing the `cpb-aacip-`/etc portion of the `id` param with `cpb-aacip?`, and searching solr once for this value.

The `/api/:id` route currently does not account for guid variations, so this same `?` fix can be applied to the api route to correctly handle guid styles on the api.

The AAPB search index fundamentally should not contain separate records for different styles of the same guid, so this fix should allow for the correct behavior across all routes with less hittin' of solr.

### Done when

API and Catalog routes handle guid variations with only one solr search required

### Additional context

_No response_

afred:
Replaces:
- `cpb-aacip-`
- `cpb-aacip/`
- `cpb-aacip_`

with: `cpb-aacip?`

so that all ID styles can be returned when querying on ID.

Closes #2773
Closes #2827

### DIFF
--- a/app/controllers/api_controller.rb
+++ b/app/controllers/api_controller.rb
@@ -37,7 +37,11 @@ class ApiController < ApplicationController
 
   def show
     @solr = Solr.instance.connect
-    data = @solr.get('select', params: { q: "id:#{params[:id]}", fl: 'xml' })
+    # Replace the delimiter at the end fo the AAPB ID prefix with a single character wildcard
+    # to return any docs that are indexed with old-style ID's, having a slash, or sometimes an underscore
+    # instead of a dash. Eventually, all records should have new-style IDs with the dashes. Until then, this.
+    id = params[:id].sub(/cpb-aacip./, "cpb-aacip?")
+    data = @solr.get('select', params: { q: "id:#{id}", fl: 'xml' })
 
     return render_not_found(params[:id]) unless data['response']['docs'] && data['response']['docs'][0]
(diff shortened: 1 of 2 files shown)


mrharpo (approved):
👍 🎉

PR #2863, opened 2024-10-15
📌 Pin CI runners to ubuntu 22.04

foglabs:
# Ubuntu
Pins the runner image to `ubuntu-22.04`

`ubuntu-latest` began failing due to missing `sqlite3.h` header. `ubuntu-22.04` builds `sqlite3` successfully, as before.


mrharpo (approved):
👍 📌

Issue #87, opened 2015-02-05
Deploy code to AWS via OpsWorks


afred:
There appears to be a few tweaks required to get it to deploy to our stack in AWS OpsWorks -- mostly adding special dependencies I think.

**done when**
Can deploy working code to AWS instance having been set up with OpsWorks
_OR_ we decide to abandon it for a regular old ec2 instance.

caseyedavis12:
@afred: Is this done?

afred:
no longer relevant

Issue #89, opened 2015-02-06
Don't load OpenSeaDragon JS?


mccalluc:
The gallery gem by default comes with a slideshow mode: I'd prefer not to do it (just another thing to design/maintain), but OpenSeaDragon is still being loaded. See if we can get rid of it so the site can be lighter.

mccalluc:
(Look at the cruft that came in with #90)

PR #2883, opened 2024-11-22
Add Demo Config, Noindex Tag When On Demo

foglabs:
We want to add a `noindex` meta tag to aapb so that demo pages don't get indexed by search engines. But, we need to know when we're demo. So, add copy of `production.rb` as `demo.rb` and use that to render the tag or not

### Dependencies:
- [ ] [openvault3_deploy #87](https://github.com/WGBH-MLA/openvault3_deploy/issues/87)
- [x] test demo deployment
- [x] => search works
- [x] => media plays
- [ ] => new `<meta>` tag is present to avoid crawling


afred (approved):
👍

afred:
@foglabs I can't get this to work for some reason. Using the branch from [PR #89 openvault3_deploy](https://github.com/WGBH-MLA/openvault3_deploy/pull/89) you can set `RAILS_ENV=demo` on deployment and that seems to work fine. However, it doesn't seem to be making it's way into `Rails.env` for the conditional, but idk why.

PR #3057, opened 2025-08-25
Update ttbook.md

michellek27:
Added links to new records, "Listening to Whales" and "The Sum of Our Data."


ekemeyer (approved):
Looks good

PR #3176, opened 2026-06-24
Add LOC_IP_RANGE to onsite_ip_ranges method

### LINKED ISSUE #3175: Add LOC back to IP Address Approved List
We removed LOC IP addresses as permitted locations for On Location items. We would like to add them back. Rachel Curtis is a good contact for testing that the update was successful

See #3083 where they were removed

ekemeyer:
Closes #3175


ekemeyer:
Confirmed that this works for Library of Congress while on call with Rachel Curtis

PR #3178, opened 2026-06-30
Private lua statement

ekemeyer:
Private items currently state you can stream them onsite at Library of Congress. However, we'd like to adjust the definition of private as not allowing streaming anywhere. The logic is already set up to not stream at LOC, but the statement needs to be updated to reflect that.


ekemeyer:
Confirming we're ready for this to go live after talking to managers on July 1, 2026.

PR #3191, opened 2026-08-10
Bot query block

ekemeyer:
Adds a block to any queries with strings greater than 1000 characters, or queries with more than 10 'OR' statements. Leads to Bad Request page.

Also sets cookie expiration to 2 hours, as opposed to 24.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -304,11 +304,11 @@ class CatalogController < ApplicationController
   end
 
   def reject_abusive_queries
-  query_string = request.query_string
+    query_string = request.query_string
     
     if query_string.length > 1000 || query_string.scan(/\bOR\b/).count > 10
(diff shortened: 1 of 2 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -302,6 +303,15 @@ def require_turnstile
     redirect_to turnstile_challenge_path(return_to: request.fullpath)
   end
 
+  def reject_abusive_queries
+    query_string = request.query_string
+
+    if query_string.length > 1000 || query_string.scan(/\bOR\b/).count > 10
+      Rails.logger.warn("Rejected abusive catalog query from #{request.remote_ip}: #{query_string.truncate(200)}")
+      render plain: "Bad Request: please submit a valid query", status: :bad_request
```
afred, line 311: This error message could definitely be more abusive and colorful

afred (approved):
👍 Looks great!

Issue #3204, opened 2026-09-01
David Grubin automatic search results


ekemeyer:
# Details 
Hey Erica, I had a call with Chris Wolf of David Grubin yesterday, and he had a question that I'm passing along here: 
when you see David Grubin's materials from his org page, it automatically sorts by "Relevance" -- is it possible to have it automatically sort by "Title" instead? (not sure how viable it is!)
He mainly asks since "Relevance" doesn't necessarily list things in order (ex: on the first page of "Relevant" results, Destination America episode 4 is listed above episode 3), since he worries about how confusing it might be for aged users and/or non-tech-savvy folks.
Idk how viable that is, but figured I'd pass along the question!
Thanks!
-Miranda

Submitted by: Miranda
CC in communications: 
Priority: Medium (within this month)
URL: https://americanarchive.org/catalog?f%5Bcontributing_organizations%5D%5B%5D=David+Grubin+Productions%2C+Inc.+%28NY%29&f[access_types][]=digitized
Slack message thread:

ekemeyer:
Not possible to force the sort from relevancy to title without creating an override or changing all organization links to be sorted by title. will discuss at Mellon 4 meeting if there's interest in changing that across the board.

ekemeyer:
merged, done!

PR #3206, opened 2026-09-02
Force org sort by title asc

ekemeyer:
See #3204, still need to discuss with AAPB team on this shift.
UPDATE: Approved


foglabs (approved):
💪

